Issue #47, opened 2019-03-07
add file load generator


6syun9:
Hi.

I want the following `load_generator` features.
```
class Hoge(gokart.TaskOnKart):
    
    ...

    def run(self):
        for x in self.load_generator('piyo'):
            proccess_file(x)
```
`gorkart` is the pipeline for data processing. I think `load_generator` is effective for sequential processing and large file that couldn't ride on memory. How do you think?
Thanks.

nishiba:
Hi. 
Thank you for your proposal. I agree with you.
Would you implement the function?

PR #48, opened 2019-03-17
add load_generator

6syun9:
https://github.com/m3dev/gokart/issues/47

usage:
```
class Hoge(gokart.TaskOnKart):

    ...

    def run(self):
        result = []
        for x in self.load_generator():
            result.append( proccess_file(x) )
       self.dump(result)
```

`load_generator` reads the divided file into iterations.
It's effective when can't read all data to memory, because `load_generator` doesn't load all files at once.


nishiba:
Good job!! Thanks!!

PR #56, opened 2019-04-11
Support json and xml extensions

kuri8ive:
Responding to the need to handle json and xml

### DIFF
--- a/gokart/file_processor.py
+++ b/gokart/file_processor.py
@@ -2,6 +2,7 @@ import pickle
 import os
 from abc import abstractmethod
 from logging import getLogger
+import xml.etree.ElementTree as ET
 
 import luigi
 import luigi.contrib.s3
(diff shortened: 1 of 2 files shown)


gokart/file_processor.py
```diff
@@ -2,6 +2,7 @@
 import os
 from abc import abstractmethod
 from logging import getLogger
+import xml.etree.ElementTree as ET
```
hirosassa: [IMO] for unity of the codes,  I recommend `et` instead of `ET` (for example, pandas as pd)

6syun9:
@kuri8ive LGTM!! Thank you great pull request!

nishiba:
LGTM

gokart/file_processor.py
kuri8ive: In this case, I followed [Official docs](https://docs.python.org/ja/3/library/xml.etree.elementtree.html). Thanks anyway!

hirosassa:
LGTM!

PR #58, opened 2019-04-12
use zipfile

nishiba:
This PR uses zipfile mdule instead of shutil, because shutil fails to unzip large files.

### DIFF
--- a/gokart/zip_client.py
+++ b/gokart/zip_client.py
@@ -1,8 +1,15 @@
 import os
 import shutil
+import zipfile
 from abc import abstractmethod
 
 from gokart.s3_config import S3Config
+
+
+def _unzip_file(filename: str, extract_dir: str) -> None:
+    zip_file = zipfile.ZipFile(filename)
+    zip_file.extractall(extract_dir)
+    zip_file.close()
 
 
 class ZipClient(object):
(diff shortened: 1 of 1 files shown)


nishiba:
@hirosassa @6syun9 Please check my codes. Thanks!

gokart/zip_client.py
```diff
@@ -62,13 +69,12 @@ def exists(self) -> bool:
 
     def make_archive(self) -> None:
         extension = os.path.splitext(self._file_path)[1]
-        shutil.make_archive(
-            base_name=self._temporary_directory, format=extension[1:], root_dir=self._temporary_directory)
+        shutil.make_archive(base_name=self._temporary_directory, format=extension[1:], root_dir=self._temporary_directory)
```
hirosassa: How about using Zipfile package for compression? For upzip, you already use Zipfile.

I think it's better using same package for simplicity. Or are there any reason for using `shutil`?

6syun9:
LGTM

gokart/zip_client.py
nishiba: It sounds good. I fix codes to use zipfile to compress files.

nishiba: @hirosassa I tried to use zipfile instead of shutil It found that code for compression with zipfile is like complicated. Please see https://stackoverflow.com/questions/1855095/how-to-create-a-zip-archive-of-a-directory . So I think it is better to use shutil to compress directories until problems occur.

hirosassa: Understood! Thanks for investigating!! LGTM

PR #63, opened 2019-05-03
add setuptools_scm for removing verup annoying procedure

hirosassa:
Since we tend to forget update version number in setup.py, I add setuptools_scm for detecting version number from git tag automatically.


nishiba:
LGTM!!

PR #71, opened 2019-07-24
caching unique id

liyuan9988:
To cache unique id once given.

### DIFF
--- a/gokart/task.py
+++ b/gokart/task.py
@@ -179,9 +179,10 @@ class TaskOnKart(luigi.Task):
         self._get_output_target(target).dump(obj)
 
     def make_unique_id(self):
-        if self.task_unique_id is not None:
-            return self.task_unique_id
+        self.task_unique_id = self.task_unique_id or self._make_unique_id()
+        return self.task_unique_id
 
+    def _make_unique_id(self):
         def _to_str_params(task):
             if isinstance(task, TaskOnKart):
                 return str(task.make_unique_id())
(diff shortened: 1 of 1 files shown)


gokart/task.py
```diff
@@ -188,6 +193,7 @@ def _to_str_params(task):
         dependencies.append(self.__class__.__name__)
         return hashlib.md5(str(dependencies).encode()).hexdigest()
 
+
```
6syun9: Useless blank lines.

gokart/task.py
```diff
@@ -178,6 +179,10 @@ def dump(self, obj, target: Union[None, str, TargetOnKart] = None) -> None:
         self._get_output_target(target).dump(obj)
 
     def make_unique_id(self):
+        self.task_unique_id = self.task_unique_id or self._make_unique_id()
+        return self.task_unique_id
+
+    def _make_unique_id(self):
```
6syun9: I think this naming is bad because it's different from the actual condition of `make_unique_id` .
Maybe, there is a good naming, for example `_make_hash` , `_make_hash_id` ,.... How do you think? @liyuan9988

gokart/task.py
liyuan9988: Thank you. I will fix it.

gokart/task.py
liyuan9988: I see. I would prefer `_make_hash_id`.
How about `self.task_unique_id`? Should we change it to `self.task_hash_id`?

6syun9: I think `task_unique_id` method does not need to be changed.

I was worried about having `task_unique_id` and `_task_unique_id`. The functions were completely different though they had similar names.

liyuan9988: fixed

gokart/task.py
liyuan9988: fixed

Issue #72, opened 2019-07-27
More parameters are required to reproduce the experiment


6syun9:
Hi. @nishiba @hirosassa 

Now, we are writing the parameter log in `~/task_param/*` as follows.
```
 {"from_date": "2019-07-04", "to_date": "2019-07-05", "type": "hoge"}
```
But, only its own parameters ,so other parameters are required to reproduce the experiment.  
We also need the parameters of the task being `luigi.task.flatten(self.requires())`.

Ideal for example...
```
 {"from_date": "2019-07-04", "to_date": "2019-07-05", "type": "hoge", "interval": 1, "item_type": "piyo", ...}
```

We should edit here!
https://github.com/m3dev/gokart/blob/master/gokart/task.py#L247

6syun9:
From #73 
> For example, I want to get the string data/piyo.csv from outside of gokart in following cases,.
It's difficult because we need to write self.make_target('data/piyo.csv') directly to the code, and make_target need require's parameter.
```
class Hoge(gokart.TaskOnKart)
    def output(self):
        return self.make_target('data/piyo.csv')
```

I want get `data/piyo.csv`...

nishiba:
When we know the task id, please give the id  to the 'make_target' as follow:
`self.make_target('data/piyo_THE_TASK_IS.csv', use_unique_id=False)`

6syun9:
@nishiba We can't get string (`data/piyo_{}_.csv`) from modulized `Hoge`.
We can see that if you look at the library source, but we usually don't want it...

6syun9:
fix #76

PR #73, opened 2019-07-27
change dump_task_params

6syun9:
before:
```
{"from_date": "2019-07-04", "to_date": "2019-07-05", "type": "hoge"}
```
after:
```
{"TaskA": {"from_date": "2019-07-04", "to_date": "2019-07-05", "type": "hoge"}, "TaskB": {"item_type": "piyo"}}
```

https://github.com/m3dev/gokart/issues/72

### DIFF
--- a/gokart/task.py
+++ b/gokart/task.py
@@ -244,7 +244,9 @@ class TaskOnKart(luigi.Task):
 
     @luigi.Task.event_handler(luigi.Event.START)
     def _dump_task_params(self):
-        self.dump(self.to_str_params(only_significant=True), self._get_task_params_target())
+        dependencies = {task.__class__.__name__: task.to_str_params(only_significant=True) for task in luigi.task.flatten(self.requires())}
+        dependencies.update({self.__class__.__name__: self.to_str_params(only_significant=True)})
+        self.dump(dependencies, self._get_task_params_target())
 
     def _get_processing_time_target(self):
         return self.make_target(f'log/processing_time/{type(self).__name__}.pkl')


nishiba:
I want to check this PR in detail.  I think this PR makes too large log file when task trees are deep.

6syun9:
@nishiba Ok. If you care about param_log file size, we can use `_to_hash(luigi.task.flatten(self.requires()))`.
I will investigate CI...

nishiba:
@6syun9 Do you know the following function?  https://github.com/m3dev/gokart/blob/89d62841362994402d178bb753a241ae9442e772/gokart/task.py#L263
We can restore any task by its hash id.

And we can check all parameters with the command:
https://gokart.readthedocs.io/en/latest/task_information.html#task-logs

6syun9:
@nishiba `restore` method create a new instance of `TaskOnKart()`.  Is this compatible with S3?
If `restore` corresponds to S3, maybe I can do what I want to do!

6syun9:
I understand the gokart ecosystem quite well. But I think It's difficult to `restore` from the outside.

For example,  I want to get the string `data/piyo.csv` from outside of gokart in following cases,.
It's difficult because we need to write `self.make_target('data/piyo.csv')` directly to the code, and `make_target` need `require`'s parameter.
```
class Hoge(gokart.TaskOnKart)
    def output(self):
        return self.make_target('data/piyo.csv')
```

Solution for `from the outside`, How about making `gokart.TaskOnKart.output_target`  parameter?
```
{"from_date": "2019-07-04", "to_date": "2019-07-05", "type": "hoge", "output_target": "data/piyo.csv"}
```
This solution can restore from `parameter.pkl` only to `data/piyo.csv`. This will widen the outside.

@nickhall How do you think?
This PR will close :)

PR #75, opened 2019-08-09
WIP: add file path log

6syun9:
https://github.com/m3dev/gokart/issues/72


6syun9:
Maybe better to use task.task_log.update(). close

PR #80, opened 2019-08-19
Add gcs support

dasoran:


dasoran:
I try to add GCS support.
There are no mock modules for GCS, so I haven't do test this code.

hirosassa:
@dasoran 
Thanks for your  effort! Looks GREAT!!
Could you add a migration guide on README to migrate existing codes to this update, if you have time?

hirosassa:
@6syun9 @nishiba ping

setup.py
```diff
@@ -12,7 +12,7 @@
     'python-dateutil==2.7.5',
     'boto3',
     'slackclient>=2.0.0',
-    'pandas',
+    'pandas<0.25.0',
```
6syun9: I think it's better to lock pandas version because there is a difference in pickle loading due to version differences.
For example, pandas 0.24.x can't load pickle made with pandas 0.23.x.
https://stackoverflow.com/questions/54665527/importerror-no-module-named-pandas-core-internals-managers-pandas-core-inte

6syun9:
Great PR! LGTM except comment!

dasoran:
> @dasoran
> Thanks for your effort! Looks GREAT!!
> Could you add a migration guide on README to migrate existing codes to this update, if you have time?

Thank you for your review :)
OK, I try to write README. wait a moment.

PR #86, opened 2019-09-02
fix slack class to read over 1000 channels

dasoran:
Because of a problem with the next_cursor part of the slack api, if the number of channels exceeds 1000, all slack channels cannot be acquired.

So I implemented a process to recursively get all the slack channels.

https://api.slack.com/methods/channels.list


nishiba:
LGTM

hirosassa:
LGTM

PR #92, opened 2019-09-17
Add log file suffix option

mski-iksm:
I've added a parameter to change the output format of log files.
Deafult format is pkl.


Teppei-Kanayama:
LGTM

nishiba:
@mski-iksm Why do you need to change output file type of log? Please try thunderbolt ( https://github.com/m3dev/thunderbolt ).

PR #93, opened 2019-09-18
do not fix pandas version

nishiba:
I think that applications which use gokart should fix pandas version.


vaaaaanquish:
LGTM

hirosassa:
LGTM

Issue #96, opened 2019-09-22
slack response don't have response_metadata


vaaaaanquish:
I found following error.
```
slack/slack_api.py", line 37, in _get_channels
    if response['response_metadata']['next_cursor']:
TypeError: 'NoneType' object is not subscriptable
```

PR #97, opened 2019-09-22
fix issue #96

vaaaaanquish:
slack response don't have response_metadata : https://github.com/m3dev/gokart/issues/96


hirosassa:
LGTM

PR #102, opened 2019-09-24
rm require pyhton-dateutil

vaaaaanquish:
We fixed python-dateutil version.  Now it is unnecessary :)
- reference 
https://github.com/spotify/luigi/issues/2662, https://github.com/spotify/luigi/blob/master/setup.py#L40


Hi-king:
LGTM

PR #107, opened 2019-10-24
fix gcsclient

hirosassa:


gokart/gcs_zip_client.py
```diff
@@ -21,8 +21,8 @@ def make_archive(self) -> None:
 
     def unpack_archive(self) -> None:
         os.makedirs(self._temporary_directory, exist_ok=True)
-        self._client.get(self._file_path, self._temporary_file_path())
-        _unzip_file(filename=self._temporary_file_path(), extract_dir=self._temporary_directory)
+        file_pointer = self._client.download(self._file_path)
+        _unzip_file(filename=file_pointer, extract_dir=self._temporary_directory)
```
dasoran: function `_unzip_file` has a type miss match.

https://docs.python.org/3/library/zipfile.html#zipfile-objects
In `_unzip_file`, the first arguments is passed to `ZipFile` constructor.
A first arguments of `ZipFile` constructor receives 'str', FilePointer or Path-like object.
So this code is correct to run but is not correct as typing.

I think you should remove the type hint of the first arguments of `_unzip_file` or create new function to receive FilePointer

dasoran:
```
diff --git a/gokart/zip_client.py b/gokart/zip_client.py
index b4b47fa..5616580 100644
--- a/gokart/zip_client.py
+++ b/gokart/zip_client.py
@@ -2,10 +2,11 @@ import os
 import shutil
 import zipfile
 from abc import abstractmethod
+from typing import Union, IO


-def _unzip_file(filename: str, extract_dir: str) -> None:
-    zip_file = zipfile.ZipFile(filename)
+def _unzip_file(fp: Union[str, IO, os.PathLike], extract_dir: str) -> None:
+    zip_file = zipfile.ZipFile(fp)
     zip_file.extractall(extract_dir)
     zip_file.close()
```

vaaaaanquish:
LGTM

PR #110, opened 2019-12-17
Add gcs discover cache

dasoran:
Now, there are many calls to Discover API. So I try to cache Discover API result.

I cannot pass test `test_get_gcs_client_without_gcs_credential_name`. Please tell me the problems.


nishiba:
LGTM

PR #113, opened 2020-01-13
add framework to run tasks with empty dataframe.

nishiba:
I would like to add a test framework to check if tasks work well with empty pandas data frame.
The usage is as follows:
```
python main.py SomeTask --local-scheduler --test-run-pandas
```

The output is as follows:
```
status=OK; namespace=test_run_with_empty_data_frame.dummy; name=DummyModelTask; id=b6c763e5b5d4041d98b600d029edbfb1;
status=OK; namespace=test_run_with_empty_data_frame; name=DummyPandasDataFrameTask; id=b15384f2a8ea02ceb5b821a655a47fcb;
status=OK; namespace=test_run_with_empty_data_frame; name=DummyWorkFlowWithoutError; id=9d7dede73411a543cca497851d5a3f7b;
```

To filter tasks by task_namespace, please add `--test-run-namespace=`.


saiya:
(Just asking) What happen if the target task has side-effect?

hirosassa:
Great work! Thanks!

gokart/testing/check_if_run_with_empty_data_frame.py
```diff
@@ -0,0 +1,81 @@
...
+        all_tasks = _get_all_tasks(cp.get_task_obj())
+
+    if test_run_params.namespace is not None:
+        all_tasks = [t for t in all_tasks if t.task_namespace == test_run_params.namespace]
+
+    with patch('gokart.TaskOnKart.load_data_frame', new=lambda *args, required_columns=None, **kwargs: pd.DataFrame(columns=required_columns)):
+        with patch('gokart.TaskOnKart.dump', new=lambda *args, **kwargs: None):
+            test_status_list = [_run_with_test_status(t) for t in all_tasks]
+
+    print('\n'.join(s.format() for s in test_status_list))
```
vaaaaanquish: I think we should use `logger.info`.
Because we may want to operate from another module.
```suggestion
    logger.info('\n'.join(s.format() for s in test_status_list))
```

vaaaaanquish:
LGTM. Thanks:)

gokart/testing/check_if_run_with_empty_data_frame.py
nishiba: @vaaaaanquish Thanks! I used logger instead of print.

nishiba:
@saiya  No problem. This test framework checks if one single task runs well with empty **pandas data frame** as input.

PR #114, opened 2020-01-19
LargeDataFrameProcessor can handle list/dict of dataframes

Hi-king:
Currently LargeDataFrameProcessor (or its inteface `TaskOnKart.make_large_data_frame_target`) supports only dataframe. 
This PR enables it to handle list/dict of dataframes as follows

```
class Task(TaskOnKart):
    def output(self):
        return self.make_large_data_frame_target('data.zip')

    def run(self):
        self.dump({
            "a": [
                pd.DataFrame(dict(data=np.random.uniform(0, 1, size=int(1e+6)))),
                pd.DataFrame(dict(data=np.random.uniform(0, 1, size=int(1e+6)))),
            ],
            "b": pd.DataFrame(dict(data=np.random.uniform(0, 1, size=int(1e+6))))
        })

```

### DIFF
--- a/gokart/target.py
+++ b/gokart/target.py
@@ -1,5 +1,6 @@
 import hashlib
 import os
+import pathlib
 import shutil
 from abc import abstractmethod
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


nishiba:
@Hi-king  Why do you not use list/dict of `self.,make_large_data_frame_target`?

PR #122, opened 2020-02-06
make default TaskOnKart.requires

nishiba:
When a task has instances of TaskInstanceParameter, then default `requires` function returns dictionary of these instances as follows:

```python
class SomeTask(gokart.TaskOnKart):
    a_task = gokart.TaskInstanceParameter()
    b_task = gokart.TaskInstanceParameter()

    # This PR makes the following function automatically.
    def requires(self):
        return dict(a=self.a_task, b=self.b_task)
```

### DIFF
--- a/gokart/task.py
+++ b/gokart/task.py
@@ -61,8 +61,11 @@ class TaskOnKart(luigi.Task):
         return self.make_target(os.path.join(file_path, f"{type(self).__name__}.pkl"))
 
     def requires(self):
-        tasks = {key[:-5]: var for key, var in vars(self).items() if key.endswith('_task') and isinstance(var, TaskOnKart)}
+        tasks = self.make_task_instance_dictionary()
         return tasks or []  # when tasks is empty dict, then this returns empty list.
+
+    def make_task_instance_dictionary(self) -> Dict[str, 'TaskOnKart']:
+        return {key[:-5]: var for key, var in vars(self).items() if key.endswith('_task') and isinstance(var, TaskOnKart)}
 
     @classmethod
     def _add_configuration(cls, kwargs, section):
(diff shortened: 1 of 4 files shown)


gokart/task.py
```diff
@@ -60,6 +60,13 @@ def output(self):
         file_path = self.__module__.replace(".", "/")
         return self.make_target(os.path.join(file_path, f"{type(self).__name__}.pkl"))
 
+    def requires(self):
+        tasks = self.make_task_instance_dictionary()
+        return tasks or []  # when tasks is empty dict, then this returns empty list.
+
+    def make_task_instance_dictionary(self) -> Dict[str, 'TaskOnKart']:
+        return {key[:-5]: var for key, var in vars(self).items() if key.endswith('_task') and isinstance(var, TaskOnKart)}
```
Hi-king: [ASK] What is the intention requiring "_task" for identifier? I think just checking `TaskOnKart ` type can do.

nishiba: I prefere
```python
SomeTask(some_data_task=SomeDataTask())
...
some_data = self.load_data_frame('some_data_task')
```
than
```python
SomeTask(some_data=SomeDataTask())
...
some_data = self.load_data_frame('some_data')
```

Hi-king: I think ignoring task without specific prefix is a little bit confusing.
How about following alternative?

```suggestion
        return {re.sub(r'_model$', '', key) var for key, var in vars(self).items() if isinstance(var, TaskOnKart)}
```

Hi-king:
Anyway, this feature should help us keep our code simple!
LGTM

nishiba:
@Hi-king I think it is confusing too. So `key` should not be changed in `requires`.

nishiba:
@Hi-king @vaaaaanquish Please review this PR.  

I change the specification about `key`. `key` is not change in the default `requires` function.

Hi-king:
@nishiba LGTM!

PR #123, opened 2020-02-07
Fix to use _LargeLocalFileReader for local (and GCS) files

Hi-king:
# Problem
Since https://github.com/m3dev/gokart/pull/80, _LargeLocalFileReader has never been used.
Because ObjectStorage. is_readable_objectstorage_instance is the flag to use _LargeLocalFileReader or not.

This problem is only appears when loading files larger than 4GB on Mac.
https://stackoverflow.com/questions/31468117/python-3-can-pickle-handle-byte-objects-larger-than-4gb 

# Workaround
I have tested that GCS client has ability to read files with chunk.
So I have changed `ObjectStorage. is_readable_objectstorage_instance` only to  prevent S3 files from using `_LargeLocalFileReader`

### DIFF
--- a/gokart/object_storage.py
+++ b/gokart/object_storage.py
@@ -10,11 +10,8 @@ from gokart.s3_zip_client import S3ZipClient
 from gokart.gcs_zip_client import GCSZipClient
 from gokart.zip_client import ZipClient
 
+object_storage_path_prefix = ['s3://', 'gs://']
 
-object_storage_path_prefix = [
-    's3://',
-    'gs://'
-]
 
 class ObjectStorage(object):
     @staticmethod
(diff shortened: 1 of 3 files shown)


gokart/object_storage.py
```diff
@@ -67,11 +61,6 @@ def get_zip_client(file_path: str, temporary_directory: str) -> ZipClient:
...
             raise
 
-
     @staticmethod
     def is_readable_objectstorage_instance(file: object):
-        if isinstance(file, luigi.contrib.s3.ReadableS3File):
-            return True
-        else:
-            # for GCS object
-            return isinstance(file, FileWrapper)
```
Hi-king: The bug was here.
This logic has prevented LocalFile from using `_LargeLocalFileReader`

nishiba: LGTM

hirosassa: Great! Thanks!

hirosassa:
LGTM

Hi-king:
@dasoran could you check this out?

PR #127, opened 2020-02-17
Add config of setting make tree info for slack notification

Ryusuketa:
I would like to add config to set making dependency tree information of gokart tasks for Slack notification.
This processing sometimes take long time to make dependency tree information.
Specifically, the larger the number of dependency tasks become, the longer the time to make this information is taken.
Therefore, I add the option to select which the tree information is sent or not to Slack notification configuration.

### DIFF
--- a/gokart/run.py
+++ b/gokart/run.py
@@ -99,7 +99,7 @@ def _try_to_send_event_summary_to_slack(slack_api: Optional[gokart.slack.SlackAP
         os.linesep,
     ])
     if options.send_tree_info:
-        content += os.linesep.join(['==== Tree Info ====', tree_info])
+        content = os.linesep.join([content, '==== Tree Info ====', tree_info])
     slack_api.send_snippet(comment=comment, title='event.txt', content=content)
 
 
--- a/gokart/run.py
+++ b/gokart/run.py
@@ -89,7 +89,7 @@ def _try_to_send_event_summary_to_slack(slack_api: Optional[gokart.slack.SlackAP
     options = gokart.slack.SlackConfig()
     with CmdlineParser.global_instance(cmdline_args) as cp:
         task = cp.get_task_obj()
-        tree_info = gokart.make_tree_info(task, details=True) if options.send_tree_info else ''
+        tree_info = gokart.make_tree_info(task, details=True) if options.send_tree_info else 'Please add SlackConfig.send_tree_info to include tree-info'
         task_name = type(task).__name__
 
     comment = f'Report of {task_name}' + os.linesep + event_aggregator.get_summary()
@@ -97,9 +97,9 @@ def _try_to_send_event_summary_to_slack(slack_api: Optional[gokart.slack.SlackAP
         '===== Event List ====',
         event_aggregator.get_event_list(),
         os.linesep,
+        '==== Tree Info ====',
+        tree_info
     ])
-    if options.send_tree_info:
-        content = os.linesep.join([content, '==== Tree Info ====', tree_info])
     slack_api.send_snippet(comment=comment, title='event.txt', content=content)
 
 
--- a/gokart/run.py
+++ b/gokart/run.py
@@ -86,9 +86,10 @@ def _try_to_send_event_summary_to_slack(slack_api: Optional[gokart.slack.SlackAP
     if slack_api is None:
         # do nothing
         return
+    options = gokart.slack.SlackConfig()
     with CmdlineParser.global_instance(cmdline_args) as cp:
         task = cp.get_task_obj()
-        tree_info = gokart.make_tree_info(task, details=True)
+        tree_info = gokart.make_tree_info(task, details=True) if options.send_tree_info else ''
         task_name = type(task).__name__
 
     comment = f'Report of {task_name}' + os.linesep + event_aggregator.get_summary()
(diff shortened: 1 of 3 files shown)


gokart/run.py
```diff
@@ -86,19 +86,20 @@ def _try_to_send_event_summary_to_slack(slack_api: Optional[gokart.slack.SlackAP
...
     comment = f'Report of {task_name}' + os.linesep + event_aggregator.get_summary()
     content = os.linesep.join([
         '===== Event List ====',
         event_aggregator.get_event_list(),
         os.linesep,
-        '==== Tree Info ====',
-        tree_info,
     ])
+    if options.send_tree_info:
+        content += os.linesep.join(['==== Tree Info ====', tree_info])
```
Hi-king: In this code, the linesep between content and tree_info might be eliminated.
How about 

```suggestion
        content = os.linesep.join([content, '==== Tree Info ====', tree_info])
```

or run `os.linesep.join` at the very end.

Hi-king: or... might there be redundant linesep in original code?

test/test_run.py
```diff
@@ -46,6 +49,35 @@ def test_run_tree_info(self):
...
+        def get_content(content: str, **kwargs):
+            self.output = content
+        slack_api_mock = MagicMock()
+        slack_api_mock.send_snippet.side_effect = get_content
+
+        cmdline_args = [f'{__name__}._DummyTask', '--param', 'test']
+        with patch('gokart.slack.SlackConfig.send_tree_info', True):
+            _try_to_send_event_summary_to_slack(slack_api_mock, event_aggregator_mock, cmdline_args)
+        expects = os.linesep.join(['===== Event List ====', event_aggregator_mock.get_event_list(), os.linesep])\
+            + os.linesep.join(['==== Tree Info ====', 'tree'])
```
Hi-king: [nits]
Following code might be simpler.
```
os.linesep.join(['===== Event List ====', event_aggregator_mock.get_event_list(), os.linesep, '==== Tree Info ====', 'tree'])
```

Hi-king:
LGTM

gokart/run.py
Ryusuketa: Thank you. I fixed it.

test/test_run.py
Ryusuketa: Thank you! I fixed it!!!!

gokart/slack/slack_config.py
```diff
@@ -5,3 +5,6 @@ class SlackConfig(luigi.Config):
     token_name = luigi.Parameter(default='SLACK_TOKEN', description='slack token environment variable.')
     channel = luigi.Parameter(default='', description='channel name for notification.')
     to_user = luigi.Parameter(default='', description='Optional; user name who is supposed to be mentioned.')
+    send_tree_info = luigi.BoolParameter(default=True,
```
nishiba: When `default=True`,  `send_tree_info` can not set False in command line.

Ryusuketa: Yeah, I see your point. But, if `default=False`, this updating is not compatible for backward versions.
I think better that this parameter is set to `False` in script file including `gokart.run()`  or `config.ini` than backward compatible is lost.

When `default=False` is adopted, to retain backward compatible, the code style maybe become ugly... (ex. `send_tree_info` is named `do_not_send_tree_info`.)

gokart/run.py
```diff
@@ -86,19 +86,20 @@ def _try_to_send_event_summary_to_slack(slack_api: Optional[gokart.slack.SlackAP
     if slack_api is None:
         # do nothing
         return
+    options = gokart.slack.SlackConfig()
     with CmdlineParser.global_instance(cmdline_args) as cp:
         task = cp.get_task_obj()
-        tree_info = gokart.make_tree_info(task, details=True)
+        tree_info = gokart.make_tree_info(task, details=True) if options.send_tree_info else ''
```
nishiba: I suggest `else "Please add SlackConfig. send_tree_info to include tree-info" `, and do not delete Line 99 & 100 in master, and delete Line 101 & 102 in this branch.

gokart/slack/slack_config.py
nishiba: @Ryusuketa I think you may not care about "backward compatible".

Ryusuketa: @nihsiba OK. I set `False` to default value.

gokart/run.py
Ryusuketa: @nishiba Thanks for your point. This statement is replaced. And changed statement around Line 99 & 100 reverted.

gokart/slack/slack_config.py
```diff
@@ -5,3 +5,6 @@ class SlackConfig(luigi.Config):
     token_name = luigi.Parameter(default='SLACK_TOKEN', description='slack token environment variable.')
     channel = luigi.Parameter(default='', description='channel name for notification.')
     to_user = luigi.Parameter(default='', description='Optional; user name who is supposed to be mentioned.')
+    send_tree_info = luigi.BoolParameter(default=False,
+                                         description='When this option is true, the dependency tree of tasks is included in send messeage.'
+                                                     'It is recommended to set false to this option when notification takes long time.')
```
vaaaaanquish: [IMO] Can {`channel`, `to_user`, `send_tree_info`} be not `significant=False`?
Parameter hash changes each time these parameters are changed. (Maybe luigi.Config doesn't matter)

Ryusuketa: Thank you. I added default values `significant=False`.

Issue #116, opened 2020-01-23
Compression random seed number


vaaaaanquish:
Machine learning algorithms use many random numbers.
For example, `np.random.seed`, `random.seed`, `torch.random`, ... , and more.
Preserving that number is very useful in Kaggle and other random seed unsumble-enabled competitions.

I think I can support this on gokart :)

PR #128, opened 2020-02-24
Fix random seed

vaaaaanquish:
Added automatic setting of seed value according to hash, and dump of seed value.
Available in competitions like Kaggle :)

issue: https://github.com/m3dev/gokart/issues/116

### DIFF
--- a/examples/sample_fix_random_seed.py
+++ b/examples/sample_fix_random_seed.py
@@ -12,11 +12,31 @@ class SampleTask(gokart.TaskOnKart):
     def run(self):
         x = [random.randint(0, 100) for _ in range(0, 10)]
         y = [np.random.randint(0, 100) for _ in range(0, 10)]
-        self.dump({'random': x, 'numpy': y})
+        try:
+            import torch
+            z = [torch.randn(1).tolist()[0] for _ in range(0, 5)]
+        except ImportError:
+            z = []
+        self.dump({'random': x, 'numpy': y, 'torch': z})
 
 
 if __name__ == '__main__':
-    # The output is as follows every time.
-    # {'random': [57, 2, 38, 65, 8, 26, 54, 14, 16, 41], 'numpy': [96, 65, 35, 38, 79, 22, 51, 56, 25, 62]}
+    # //----------------------------------------------------------------
+    # Please set fix_random_seed_methods parameter.
     # Change seed if you change sample_param.
-    gokart.run(['sample_fix_random_seed.SampleTask', '--local-scheduler', '--rerun', '--sample-param=a'])
+    #
+    # //--- The output is as follows every time (installed pytorch). ---
+    # {'random': [60, 27, 7, 56, 91, 38, 72, 37, 69, 51],
+    #   'numpy': [20, 17, 22, 65, 87, 4, 30, 76, 81, 31],
+    #   'torch': [0.4886833131313324, 0.8920509815216064,
+    #     0.6191630959510803, -0.13227537274360657, -0.6677181720733643]}
+    #
+    # //--------------------- Not install pytorch ----------------------
+    # {'random': [60, 27, 7, 56, 91, 38, 72, 37, 69, 51],
+    #  'numpy': [20, 17, 22, 65, 87, 4, 30, 76, 81, 31],
+    #  'torch': []}
+    # //----------------------------------------------------------------
+    gokart.run([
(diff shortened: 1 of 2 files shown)


gokart/task.py
```diff
@@ -279,6 +281,34 @@ def get_task_params(self) -> Dict:
...
+        }
+        random.seed(random_seed)
+        np.random.seed(random_seed)
+        try:
+            import torch
+            torch.manual_seed(random_seed)
+            seeds['seed_method'].append('torch.manual_seed')
+            torch.cuda.manual_seed(random_seed)
+            seeds['seed_method'].append('torch.cuda.manual_seed')
+        except Exception:
```
nishiba: I think ImportError should be used. How do you think?

nishiba: How do I add other modules in my local project?  Do I need to add codes to gokart?

vaaaaanquish: Good comment.  
I found two issues in response to your point.

1. add other modules in local project (nishiba-san say)
2. misleading if we want to fix the seed from outside

Change to WIP once.

vaaaaanquish:
@nishiba @Hi-king Please check PR.

I add `fix_random_seed_methods` parameter.
This parameter will enter `fix seed method` list.
For example...
```
fix_random_seed_methods=["random.seed", "torch.random.manual_seed"]
```
If have not installed module in fix_random_seed_methods, that will be `pass`.

gokart/task.py
vaaaaanquish: I add `fix_random_seed_methods` :)

gokart/task.py
```diff
@@ -279,6 +280,36 @@ def get_task_params(self) -> Dict:
...
+                m(random_seed)
+                success_methods.append(method_name)
+            except ModuleNotFoundError:
+                pass
+            except AttributeError:
+                pass
+        return success_methods
+
+    def _get_random_seed(self):
+        return int(self.make_unique_id(), 16) % (2**32 - 1)  # maximum numpy.random.seed
```
Hi-king: 🆒

examples/sample_fix_random_seed.py
```diff
@@ -0,0 +1,42 @@
...
+    #     0.6191630959510803, -0.13227537274360657, -0.6677181720733643]}
+    #
+    # //--------------------- Not install pytorch ----------------------
+    # {'random': [60, 27, 7, 56, 91, 38, 72, 37, 69, 51],
+    #  'numpy': [20, 17, 22, 65, 87, 4, 30, 76, 81, 31],
+    #  'torch': []}
+    # //----------------------------------------------------------------
+    gokart.run([
+        'sample_fix_random_seed.SampleTask', '--local-scheduler', '--rerun', '--sample-param=a',
+        '--fix-random-seed-methods=["random.seed","numpy.random.seed","torch.random.manual_seed"]'
```
Hi-king: How about setting these methods as default random seed methods?

gokart/task.py
```diff
@@ -279,6 +280,36 @@ def get_task_params(self) -> Dict:
             return self.load(target)
         return dict()
 
+    @luigi.Task.event_handler(luigi.Event.START)
+    def _set_random_seed(self):
+        random_seed = self._get_random_seed()
+        seed_methods = self.try_set_seed(self.fix_random_seed_methods, random_seed)
+        self.dump({'seed': random_seed, 'seed_methods': seed_methods}, self._get_random_seeds_target())
```
Hi-king: Cool way to reproduce results.

IMO, there must be another way to inject random seeds because  task id occasionally changes.
How about introducing another params like `fix_random_seed_value`?

examples/sample_fix_random_seed.py
Hi-king: Sorry, I overlooked `default=['random.seed', 'numpy.random.seed']`. Cool!

gokart/task.py
vaaaaanquish: That's right. It looks like a `fixed task_id`. I'll try :)

vaaaaanquish: done https://github.com/m3dev/gokart/pull/128/commits/068c50819ffa0b401e1be8451b84d3f71715fc64

Teppei-Kanayama:
@vaaaaanquish LGTM!

gokart/task.py
vaaaaanquish: @Hi-king please check PR!

Hi-king: @vaaaaanquish cool way!

PR #131, opened 2020-06-21
Feature Request: BoolParameter which must be explicitly specified  value

e-mon:
`luigi.BoolParameter` already has "explicit parsing" feature, but also still has implicit behavior like follows.

```shell
$ python main.py Task --param
# param will be set as True
$ python main.py Task
# param will be set as False
```

I think my PR feature is not "must-have" for Luigi, but it's "good-to-have" feature for gokart.

### DIFF
--- a/test/test_explicit_bool_parameter.py
+++ b/test/test_explicit_bool_parameter.py
@@ -9,14 +9,9 @@ from luigi.cmdline_parser import CmdlineParser
 import gokart
 
 
-def run_locally(args):
-    temp = CmdlineParser._instance
-    try:
-        CmdlineParser._instance = None
-        run_exit_status = luigi.run(['--local-scheduler', '--no-lock'] + args)
-    finally:
-        CmdlineParser._instance = temp
-    return run_exit_status
+def in_parse(cmds, deferred_computation):
+    with CmdlineParser.global_instance(cmds) as cp:
+        deferred_computation(cp.get_task_obj())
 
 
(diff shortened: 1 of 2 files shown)


test/test_explicit_bool_parameter.py
```diff
@@ -0,0 +1,63 @@
...
+        ExplicitParsing._param = self.param
+
+
+class TestExplicitBoolParameter(unittest.TestCase):
+    def test_bool_default(self):
+        self.assertTrue(WithDefaultTrue().param)
+        self.assertFalse(WithDefaultFalse().param)
+
+    def test_parse_param(self):
+        run_locally(['ExplicitParsing', '--param', 'true'])
```
Hi-king, line 43: [IMO] As like this way (https://github.com/spotify/luigi/blob/master/test/dict_parameter_test.py#L49), Parser can be tested without  run. (Or unittest should be run without runnning CLI)

vaaaaanquish:
@e-mon Thx. LGTM.

test/test_explicit_bool_parameter.py
e-mon: Thx for your comments!
I rewrite code to use `in_parse` instead of `run_loccaly`.

Hi-king: 🆒

Hi-king:
@e-mon I've merged this. Thank you for your contribution.

PR #133, opened 2020-07-28
Fix evaluation order:  config file follows cmdline

Hi-king:
#6 have broken evaluation order of luigi params (https://luigi.readthedocs.io/en/stable/parameters.html#parameter-resolution-order).

### DIFF
--- a/gokart/task.py
+++ b/gokart/task.py
@@ -49,7 +49,6 @@ class TaskOnKart(luigi.Task):
     fix_random_seed_value = luigi.IntParameter(default=None, description='Fix random seed method value.', significant=False)
 
     def __init__(self, *args, **kwargs):
-        self._add_configuration(kwargs, self.get_task_family())
         self._add_configuration(kwargs, 'TaskOnKart')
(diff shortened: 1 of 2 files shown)


gokart/task.py
```diff
@@ -49,7 +49,6 @@ class TaskOnKart(luigi.Task):
     fix_random_seed_value = luigi.IntParameter(default=None, description='Fix random seed method value.', significant=False)
 
     def __init__(self, *args, **kwargs):
-        self._add_configuration(kwargs, self.get_task_family())
```
Hi-king, line 52: I think just introducing TaskOnKart.xxx (following line) will demand #6, and removed this line. Please let me know if this will break any other situations.

hirosassa: I agrre with your suggestion. LGTM!

vaaaaanquish:
LGTM 👍

Hi-king:
@hirosassa @vaaaaanquish Thank you for your reviews!

PR #136, opened 2020-08-26
Change gokart.gcs_config.GCSConfig's behaviour as a singleton

Hi-king:
Initializing GCSConfig is a mess process because of 
1. simply time consuming
2. noisy warning when using local default credentials (UserWarning: Your application has authenticated using end user credentials  ...)

We actually need to initialize GCSConfig just once in a whole process. So I've changed GCSConfig's behaviour as like singleton pattern.

### DIFF
--- a/gokart/gcs_config.py
+++ b/gokart/gcs_config.py
@@ -13,9 +13,9 @@ from googleapiclient.discovery import _retrieve_discovery_doc
 
 
 class GCSConfig(luigi.Config):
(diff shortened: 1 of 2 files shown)


vaaaaanquish:
thx, LGTM.

hirosassa:
Good improvement! Thanks 👍

Hi-king:
@vaaaaanquish @hirosassa Thx!

PR #137, opened 2020-08-30
Fixed load dataframe bug

mski-iksm:
[What I did]
- Added `len(data.columns) == 0` condition, when considering input dataframe is empty or not (gokart/task.py L.200)

[Why]
- When loaded dataframe (which is `data` at `load_data_frame()`) has no rows but with some columns, empty dataframe with only `required_columns` columns will be returned, which may have some missing columns

Please review.

### DIFF
--- a/gokart/task.py
+++ b/gokart/task.py
@@ -186,19 +186,18 @@ class TaskOnKart(luigi.Task):
 
         return _load(self._get_input_targets(target))
 
-    def load_data_frame(self,
-                        target: Union[None, str, TargetOnKart] = None,
-                        required_columns: Optional[Set[str]] = None,
+    def load_data_frame(self, target: Union[None, str, TargetOnKart] = None, required_columns: Optional[Set[str]] = None,
                         drop_columns: bool = False) -> pd.DataFrame:
         def _flatten_recursively(dfs):
             if isinstance(dfs, list):
                 return pd.concat([_flatten_recursively(df) for df in dfs])
             else:
                 return dfs
+
         data = _flatten_recursively(self.load(target=target))
 
         required_columns = required_columns or set()
-        if data.empty and len(data.index) == 0:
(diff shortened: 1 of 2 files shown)


gokart/task.py
```diff
@@ -186,19 +186,18 @@ def _load(targets):
...
             if isinstance(dfs, list):
                 return pd.concat([_flatten_recursively(df) for df in dfs])
             else:
                 return dfs
+
         data = _flatten_recursively(self.load(target=target))
 
         required_columns = required_columns or set()
-        if data.empty and len(data.index) == 0:
+        if data.empty and len(data.index) == 0 and len(data.columns) == 0:
```
Hi-king, line 200: [IMO] cool changes! len(required_columns - set(data.columns)) > 0 might be better ?

Hi-king (approved):

gokart/task.py
mski-iksm: I think the suggested code looks better. I applied the suggestion. Thank you.

test/test_task_on_kart.py
```diff
@@ -295,12 +293,21 @@ def test_load_data_frame_drop_columns(self):
...
+        self.assertIsInstance(df, pd.DataFrame)
+        self.assertEqual(0, df.shape[0])
+        self.assertSetEqual({'a', 'b', 'c'}, set(df.columns))
+
     def test_load_index_only_dataframe(self):
         task = _DummyTask()
         task.load = MagicMock(return_value=pd.DataFrame(index=range(3)))
 
         # connnot load index only frame with required_columns
-        self.assertRaises(AssertionError, lambda : task.load_data_frame(required_columns={'a', 'c'}))
```
Hi-king, line 303: 👍

Hi-king (approved):
LGTM! thanks @mski-iksm

PR #138, opened 2020-09-06
Fixed test bug at test/test_restore_task_by_id.py

mski-iksm:
I believe there is a bug at `test_restore_task_by_id.py`, although the test passes.

1. TaskInstanceParameter is not called implicitly, when requires() is implemented at task.
At the current code, dependent TaskInstanceParameters are only called implicitly when requires() is not implemented.
(This situation will be fixed in different PR)
Based on above, when requires() is implemented at _DummyTask on L.19, I think `sub_task = gokart.TaskInstanceParameter()` is not called.
So, deleted the `requires()` at _DummyTask.

2. There is not run() and dump() at _SubDummyTask
This causes runtime error at the luigi.build() on L.36.
So, I added `run()` method.

Please review the code.

### DIFF
--- a/gokart/task.py
+++ b/gokart/task.py
@@ -197,7 +197,7 @@ class TaskOnKart(luigi.Task):
         data = _flatten_recursively(self.load(target=target))
 
         required_columns = required_columns or set()
-        if data.empty and len(data.index) == 0 and len(data.columns) == 0:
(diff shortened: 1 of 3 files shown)


vaaaaanquish:
LGTM

Hi-king:
@mski-iksm LGTM! thx!

PR #139, opened 2020-09-06
Fix returning instance parameter at requires

mski-iksm:
[What is this PR?]
TaskInstanceParameters will be automatically added to `requires()`'s return.

[Why?]
At the current code, TaskInstanceParameters will not be returned automatically when `requires()` is overrided at task.

Please review.

### DIFF
--- a/gokart/task.py
+++ b/gokart/task.py
@@ -197,7 +197,7 @@ class TaskOnKart(luigi.Task):
         data = _flatten_recursively(self.load(target=target))
 
         required_columns = required_columns or set()
-        if data.empty and len(data.index) == 0 and len(data.columns) == 0:
+        if data.empty and len(data.index) == 0 and len(required_columns - set(data.columns)) > 0 == 0:
             return pd.DataFrame(columns=required_columns)
         assert required_columns.issubset(set(data.columns)), f'data must have columns {required_columns}, but actually have only {data.columns}.'
         if drop_columns:
--- a/gokart/task.py
+++ b/gokart/task.py
@@ -186,19 +186,18 @@ class TaskOnKart(luigi.Task):
 
         return _load(self._get_input_targets(target))
 
-    def load_data_frame(self,
-                        target: Union[None, str, TargetOnKart] = None,
-                        required_columns: Optional[Set[str]] = None,
+    def load_data_frame(self, target: Union[None, str, TargetOnKart] = None, required_columns: Optional[Set[str]] = None,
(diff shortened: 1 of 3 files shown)


vaaaaanquish:
@mski-iksm What kind of case do you have in mind?
I don't think there's a case to be made for `requires()` override.

gokart/task.py
```diff
@@ -55,14 +55,38 @@ def __init__(self, *args, **kwargs):
...
+        self.requires = self._add_task_instance(self.requires)
 
     def output(self):
         file_path = self.__module__.replace(".", "/")
         return self.make_target(os.path.join(file_path, f"{type(self).__name__}.pkl"))
 
     def requires(self):
-        tasks = self.make_task_instance_dictionary()
-        return tasks or []  # when tasks is empty dict, then this returns empty list.
+        return {}
```
Hi-king, line 65: @mski-iksm I think this implementation is a little bit tricky and confusing because of adding requirements from outside of overridden requires().

If you need that feature, you can explicitly write requires() as follows.
```
def OvrerrideTask(TaskOnKart):
   def requires(self):
     super().requires() + dict(other_task=...)
``` 
or
```
def OvrerrideTask(TaskOnKart):
   def requires(self):
     self.make_task_instance_dictionary() + dict(other_task=...)
```

Hi-king (approved):

gokart/task.py
Hi-king: @mski-iksm I think this implementation is a little bit tricky and confusing because of adding requirements from outside of overridden requires().

If you need that feature, you can explicitly write requires() as follows.
```python
def OvrerrideTask(TaskOnKart):
   def requires(self):
     super().requires() + dict(other_task=...)
``` 
or
```python
def OvrerrideTask(TaskOnKart):
   def requires(self):
     self.make_task_instance_dictionary() + dict(other_task=...)
```

Hi-king (changes requested):

mski-iksm:
Thank you for the comments.

As you've mentioned, returning parameters which is not required can cause problems.
e.g. 

```
self.param = luigi.Parameter()
self.hoge_task = gokart.TaskInstanceParameter()
self.fuga_task = gokart.TaskInstanceParameter()

def requires(self):
    if self.param == 'a':
        return self.hoge_task
    return self.fuga_task
```
(thx @vaaaaanquish )

I also think we should explicitly write requires() when parameters are needed.
So, I will close this PR.

PR #143, opened 2020-09-21
Handle task cache collision

mski-iksm:
[What is this PR?]
- Adding function to handle task cache collision

[Why?]
- When multiple applications run  same task at same workspace simultaneously, task cache may collide.
- example
  - Consider situation with same task AA running at application A and application B.
  - In such case, when one task loads the cache while the other is writing, unexpected error may happen.

[How to overcome the problem?]
- Implement cache file lock using redis
  - When gokart write/load/delete cache file, gokart will ask redis to lock file.
  - When lock is taken by other application, gokart will not run write/load/delete and wait for the lock to be released.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,3 +96,34 @@ def requires(self):
 def run(self):
     data = self.load_data_frame(required_columns={'id', 'name'})  
 ```
+
+## Advanced
+### Using task lock
+Task lock is implemented to prevent task cahche collision.
+(Originally, task cache collision may occur when same task with same parameters run at different applications parallelly.) 
+
+1. Set up a redis server at somewhere accessible from gokart/luigi jobs.
+
+    Following will run redis at your localhost.
+
+    ```bash
+    $ redis-server
+    ```
+
+2. Set redis server hostname and port number as parameters to gokart.TaskOnKart().
+
+    You can set it by adding `--redis-host=[your-redis-localhost] --redis-port=[redis-port-number]` options to gokart python script.
+
+    e.g.
+    ```bash
+    
+    python main.py sample.SomeTask --local-scheduler --redis-host=localhost --redis-port=6379
+    ```
+
+    Alternatively, you may set parameters at config file.
+
+    ```conf.ini
+    [TaskOnKart]
+    redis_host=localhost
+    redis_port=6379
+    ```
--- a/README.md
+++ b/README.md
@@ -99,6 +99,10 @@ def run(self):
 
 ## Advanced
 ### Using task lock
+#### Require
+You need to install (redis)[https://redis.io/topics/quickstart] for this advanced function.
+
+#### Description
 Task lock is implemented to prevent task cahche collision.
 (Originally, task cache collision may occur when same task with same parameters run at different applications parallelly.) 
 
--- a/gokart/redis_lock.py
+++ b/gokart/redis_lock.py
@@ -1,18 +1,21 @@
 import os
 from logging import getLogger
-from typing import NamedTuple, Optional
+from typing import NamedTuple
 
 import redis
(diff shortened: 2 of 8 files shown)


gokart/target.py
```diff
@@ -16,9 +16,17 @@
 from gokart.zip_client_util import make_zip_client
 from gokart.object_storage import ObjectStorage
 
+from gokart.redis_lock import RedisParams, make_redis_params, with_lock
+
 logger = getLogger(__name__)
 
 
+def _decorate_file_access_functions_with_redis_lock(super, self, redis_params):
+    self.load = with_lock(func=super.load, redis_params=redis_params)
```
Hi-king, line 25: @mski-iksm 

I think it better adding lock at function definition
https://github.com/m3dev/gokart/blob/61d59a02fa42fcd17fc607f4d081659efdcef8d5/gokart/target.py#L35
than replacing the method from outside of the class.

```python
class TargetOnKart(luigi.Target):
    @with_lock
    def load(self) -> Any:
        return self._load()
```

Hi-king (changes requested):

gokart/target.py
Hi-king: Or it can be clear if you write with_lock as TargetOnKart's method 

```python
class TargetOnKart(luigi.Target):
    def _with_lock(self, func):
        return with_lock(func, self.redis_params)

    def load(self) -> Any:
        return self._with_lock(self._load())
```

gokart/target.py
```diff
@@ -24,20 +26,23 @@ def exists(self) -> bool:
...
+        return self._with_lock(self._remove)()
 
     def last_modification_time(self) -> datetime:
         return self._last_modification_time()
 
     def path(self) -> str:
         return self._path()
 
+    def _with_lock(self, func):
+        return with_lock(func=func, redis_params=self._redis_params)
```
Hi-king, line 44: @mski-iksm [IMO] what happens when self._redis_params is not set? Plz care that situation :)

Hi-king: Maybe default param of `self._redis_params` should be set as None?

Hi-king: 👍

Hi-king (approved):

gokart/target.py
mski-iksm: @Hi-king 
`self._redis_params` must be an instance of `RedisParams`.

To care such cases, I'll set a default value at  `self._redis_params`.

hirosassa:
LGTM! Thanks!

README.md
```diff
@@ -95,4 +95,35 @@ def requires(self):
 
 def run(self):
     data = self.load_data_frame(required_columns={'id', 'name'})  
-```
\ No newline at end of file
+```
+
+## Advanced
+### Using task lock
```
vaaaaanquish, line 101: Please note that redis installation is required in this field.  
example:
```
### require
You need to install (redis)[link] for this advanced...
```

gokart/redis_lock.py
```diff
@@ -0,0 +1,77 @@
...
+        return self._redis_client
+
+
+def with_lock(func, redis_params: RedisParams):
+    if not redis_params.should_redis_lock:
+        return func
+
+    def wrapper(*args, **kwargs):
+        redis_client = RedisClient(host=redis_params.redis_host, port=redis_params.redis_port).get_redis_client()
+        redis_lock = redis.lock.Lock(redis=redis_client, name=redis_params.redis_key, timeout=30, blocking=True, thread_local=False)
```
vaaaaanquish, line 43: I think `timeout` is important parameter.
It's better to be able to set it up externally.

Hi-king:
@mski-iksm Merged. Thanks for adding great feature!

Issue #144, opened 2020-10-26
don't remove task log, params


vaaaaanquish:
When a task fails, `task_params` and `task_logs` are saved.

- when luigi.Event.Start task_param save :https://github.com/m3dev/gokart/blob/0b710d377093f2cf48d995d821b4ad1b4b65f635/gokart/task.py#L309

PR #145, opened 2020-10-26
[WIP]Change dump log timing

vaaaaanquish:
https://github.com/m3dev/gokart/issues/144

### DIFF
--- a/gokart/task.py
+++ b/gokart/task.py
@@ -274,7 +274,7 @@ class TaskOnKart(luigi.Task):
             return self.load(target)
         return dict()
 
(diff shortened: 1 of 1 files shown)


Hi-king:
@vaaaaanquish Don't us need failed task logs? For example, checking task_params of failed tasks.

vaaaaanquish:
@Hi-king I thought.

If we want to experiment at hand we may need.
but if we are managing the experiment in the production GCS, S3, even access to its log files will be difficult (because file amount).

I think Optional is better. How do you think?
(Temporarily change to WIP)

PR #159, opened 2020-12-30
WIP: Migrating to GitHub Actions and apply yapf on repo

hirosassa:
## What is this PR

I implemented followings;
1. GitHub Actions workflows of test and formatting
2. applied yapf

## todo

- implement publish workflow: need to consider how to use secrets


Hi-king:
@hirosassa since yapf makes piles of diffs, I recommend split PR into 1) only yapf 2) others!

setup.py
```diff
@@ -9,7 +9,7 @@
     'luigi',
     'boto3',
     'slack_sdk',
-    'pandas',
+    'pandas<1.2',
```
vaaaaanquish, line 12: [Ask] Why fix version?

hirosassa: This is problem caused by breaking change of pandas 1.2 (RO on 26 Dec.).
When we use `dump`, it will be failed with following output.

I'll fix this on next PR.

```
TypeError: write() argument must be str, not bytes
```

hirosassa:
@Hi-king Thanks for your suggestion! I'll split this PR into two parts.

PR #176, opened 2021-02-11
New feature: dump task info

mski-iksm:
Added a new feature dumping task cache information file after running tasks.

Task cache information file is a pandas dataframe which has information such as task name, unique id, cache file path, parameter and processing time of completed tasks.

Please review.
@Hi-king @hirosassa @vaaaaanquish @e-mon


gokart/run.py
```diff
@@ -118,5 +127,6 @@ def run(cmdline_args=None, set_retcode=True):
         event_aggregator.set_handlers()
         luigi.cmdline.luigi_run(cmdline_args)
     except SystemExit as e:
+        _dump_task_info(cmdline_args=cmdline_args, dump_task_info_path=dump_task_info_path)
```
vaaaaanquish, line 130: [ASK] Why implement it in such a way that you can change it when you do `run` it?

It could be an environment variable like `Slack API`.
Or, you can choose not to let the user choose like `other log files`.

gokart/run.py
```diff
@@ -1,4 +1,5 @@
 import configparser
+from gokart.target import make_target
```
vaaaaanquish, line 2: The location is not appropriate.

gokart/task_info.py
```diff
@@ -0,0 +1,48 @@
+from configparser import ConfigParser
+from logging import getLogger
+import os
+from typing import List, NamedTuple, Set
+import luigi
+from luigi.cmdline_parser import CmdlineParser
+import pandas as pd
+import gokart
```
vaaaaanquish, line 8: Please align the order. `import`

gokart/task_info.py
```diff
@@ -0,0 +1,48 @@
...
+
+class TaskInfo(NamedTuple):
+    name: str
+    unique_id: str
+    path: str
+    params: dict
+    processing_time: str
+
+
+def make_tasks_info(task: gokart.TaskOnKart, cache: Set[str], with_logging: bool = False, with_unique_id: bool = False):
```
vaaaaanquish, line 21: There is no `type hint` for return

vaaaaanquish (changes requested):

vaaaaanquish:
Add the appropriate test code.

vaaaaanquish:
@mski-iksm Thanks, so cool PR. It will be very useful during development. Please address some of the points raised.

README.md
```diff
@@ -36,6 +36,9 @@ class BasicTask(gokart.TaskOnKart):
 
         # save results with the file path {self.workspace_directory}/basic_task_{unique_id}.csv
         self.dump(results)
+
+if __name__ == '__main__':
```
Hi-king, line 40: [nits] You can remove this line to simplify the script

gokart/run.py
```diff
@@ -96,7 +98,14 @@ def _try_to_send_event_summary_to_slack(slack_api: Optional[gokart.slack.SlackAP
...
-def run(cmdline_args=None, set_retcode=True):
+def _dump_task_info(cmdline_args: List[str], dump_task_info_path: Optional[str]):
+    if dump_task_info_path is not None:
+        result, unique_id = get_task_info(cmdline_args=cmdline_args)
+        task_info_target = make_target(file_path=dump_task_info_path, unique_id=unique_id, processor=None, redis_params=None)
+        task_info_target.dump(obj=result, lock_at_dump=False)
+
+
+def run(cmdline_args=None, set_retcode=True, dump_task_info_path: Optional[str] = None):
     cmdline_args = cmdline_args or sys.argv[1:]
```
Hi-king, line 109: [IMO] You can get `dump_task_info_path` from `cmdline_args` so that we can use `dump_task_info_path` as usual way as other luigi options

Hi-king (changes requested):
Great feature! Let us improve this feature with little more conversations 🙏

gokart/task_info.py
hirosassa: We should add import order linter, IMO

mski-iksm: > We should add import order linter, IMO

Yes, I think so too. I'll add `isort` at another PR.

gokart/run.py
mski-iksm: - I thought setting environment variable is difficult, since it may cause users to make mistake or forget at settings
- I thought allowing to change save path would be kind to users
- I thought setting path at `gokart.run()` is understandable, since the dumped file contains resulted information of `run()`

https://github.com/m3dev/gokart/pull/176#discussion_r574907617
As mentioned here, I think setting dump/no dump in the `cmdline_args` is better.

I also think that there should be default file path for convenience. 

Considering above, I will change the implementation as,
- set dump/no dump in the `cmdline_args`
- dump file path can be set
- when no path is specified, the file will be dumped to default path (e.g. workspace/task_info.csv)

How do you think? @vaaaaanquish @Hi-king

gokart/task_info.py
Hi-king: cool! I can't wait to use isort!

gokart/task_info.py
```diff
@@ -0,0 +1,51 @@
...
+
+class TaskInfo(NamedTuple):
+    name: str
+    unique_id: str
+    path: str
+    params: dict
+    processing_time: str
+
+
+def make_tasks_info(task: gokart.TaskOnKart, cache: Set[str], with_logging: bool = False) -> List[TaskInfo]:
```
Hi-king, line 21: [nits] We should keep consistency: tasks_info or task_info :)

gokart/run.py
vaaaaanquish: > cmdline_args  

I feel not happy with anything that can't be managed in a `Config File`. It would be nice if we could pass the Config file.

> dump file path can be set  

LGTM

> when no path is specified, the file will be dumped to default path (e.g. workspace/task_info.csv)  

Sounds good. Something like `workspace/task_info/run_xxxxxxxxx.pkl`

PR #183, opened 2021-04-02
Add gokart build method

vaaaaanquish:
example
```python
import luigi
import gokart

gokart.add_config('./conf.ini')


class TaskA(gokart.TaskOnKart):
    x = luigi.Parameter()
    
    def run(self):
        self.dump(self.x)


class TaskB(gokart.TaskOnKart):
    t = gokart.TaskInstanceParameter()
    x = luigi.Parameter()
    
    def run(self):
        t = self.load('t')
        self.dump(','.join([t, self.x]))

task = TaskB(t=TaskA(x='hello'), x='world')
output = gokart.build(task)

print(output)    # hello, world
```


hirosassa (approved):
LGTM!

vaaaaanquish:
@hirosassa thx.

@Hi-king @mski-iksm plz review!

gokart/build.py
```diff
@@ -0,0 +1,59 @@
...
+    luigi.configuration.core.PARSER = ext
+    assert luigi.configuration.add_config_path(file_path)
+
+
+def build(task: TaskOnKart, verbose: bool = False, return_value: bool = True, reset_register: bool = True) -> Optional[Any]:
+    """
+    Run gokart task for local interpreter.
+    """
+    if reset_register:
+        luigi.task_register.Register._reg = [x for x in luigi.task_register.Register._reg
```
Hi-king, line 54: @vaaaaanquish 
[IMO] It would be better to separate this to public method. Because we can use this in other situations.

Hi-king: like `def reset_register(keep = {'gokart', 'luigi'})`

gokart/build.py
```diff
@@ -0,0 +1,59 @@
...
+                                             if x.__module__.split('.')[0] in {'gokart', 'luigi'}]  # avoid TaskClassAmbigiousException
+    _read_environ()
```
Hi-king, line 56: [ASK]
Why don't we use gokart.run._read_environ() ?

Hi-king: [ASK]
Is it better to add _check_config as like gokart.run() ?
https://github.com/m3dev/gokart/blob/7056c10cf4577d7008b5744c5cf127da5b65c849/gokart/run.py#L110

Hi-king: [ASK]
Is there a chance to call gokart.build() from gokart.run() ?
Cause there are many overlapping functions.

vaaaaanquish: All right

Hi-king (changes requested):
My main concerns are about overlappings between gokart.run and gokart.build

gokart/build.py
vaaaaanquish: It has an underscore, so I instinctively avoided it. `gokart.run._check_config` is good too.

vaaaaanquish:
@Hi-king 
> My main concerns are about overlappings between gokart.run and gokart.build

I don't think there's usually a use case for that. So, I'll write some documentation.

gokart/build.py
vaaaaanquish: > Is there a chance to call gokart.build() from gokart.run() ?

Not supposed to. These are both `endpoints`.

gokart/build.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def _get_output(task: TaskOnKart) -> Any:
+    output = task.output()
+    if type(output) == list:
+        return [x.load() for x in output]
+    return output.load()
+
+
+def add_config(file_path: str):
```
Hi-king, line 37: @vaaaaanquish 
[ASK]
Is this method should be in `build` package?
It seems this is just a helper not necessarily required for gokart.build.

gokart/build.py
```diff
@@ -0,0 +1,58 @@
...
+
+def build(task: TaskOnKart, verbose: bool = False, return_value: bool = True, reset_register: bool = True) -> Optional[Any]:
+    """
+    Run gokart task for local interpreter.
+    """
+    if reset_register:
+        _reset_register()
+    _read_environ()
+    _check_config()
+    with HideLogger(verbose):
```
Hi-king, line 56: cool way :)

Hi-king:
> I don't think there's usually a use case for that. So, I'll write some documentation.

My point is not for usage, but for developing from the point of view of the reusability. 
Anyway  current implementation is more reusable by importing functions from gokart.run :)

Hi-king (approved):

gokart/build.py
vaaaaanquish: Yes, I created `utils.py` and moved some things around.

Hi-king:
@vaaaaanquish cool refactoring. Merged. Thx :)

PR #190, opened 2021-04-09
Specific load

### LINKED ISSUE #189: Easy load specific data
```
class A(gokart.TaskOnKart):
    task = gokart.TaskInstanceParameter()

    def run(self):
        a = self.load()
```

This is uniquely determined, but Not implemented.

vaaaaanquish:
fix #189

### DIFF
--- a/gokart/task.py
+++ b/gokart/task.py
@@ -203,12 +203,13 @@ class TaskOnKart(luigi.Task):
             if isinstance(targets, list) or isinstance(targets, tuple):
                 return [_load(t) for t in targets]
             if isinstance(targets, dict):
-                if len(targets) == 1:
-                    return _load(list(targets.values())[0])
                 return {k: _load(t) for k, t in targets.items()}
             return targets.load()
 
-        return _load(self._get_input_targets(target))
+        data = _load(self._get_input_targets(target))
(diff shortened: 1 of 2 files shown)


vaaaaanquish:
@Hi-king @hirosassa Please review

mski-iksm:
@vaaaaanquish 
[IMO] I believe this PR allows `self.load()` without setting any parameter.

However, I thought we should select target name explicitly as in the following example. (I think this is possible in the current code)
If not, we have to add parameter to `self.load()` as `self.load('task1')` when we decided to add another `gokart.TaskInstanceParameter()` afterward, which may be difficult to notice.

```
class A(gokart.TaskOnKart):
    task = gokart.TaskInstanceParameter()

    def run(self):
        a = self.load('target')
```

vaaaaanquish:
@mski-iksm Thanks for the good point.

It would be good to add a condition to check if target is in data. 
```
        if isinstance(data, dict) and len(data) == 1 and target in data:
            return list(data.values())[0]
```

hirosassa (approved):
Looks good!

mski-iksm:
@vaaaaanquish LGTM

PR #193, opened 2021-04-14
Add parameter alias to inherits_config_params

mski-iksm:
This update allows to replace parameter name of inherited config Task, when using `@inherits_config_params` decorator.

Currently, when using `@inherits_config_params`, parameter name of inheriting config Task and decorated Task must be same. After this PR, parameter name can be replaced by passing parameter names dictionary `param_config2task` to `@inherits_config_params` decorator.

Please review!
@hirosassa @vaaaaanquish @e-mon @Hi-king

### DIFF
--- a/gokart/config_params.py
+++ b/gokart/config_params.py
@@ -6,17 +6,17 @@ import gokart
 
 
 class inherits_config_params:
-    def __init__(self, config_class: luigi.Config, param_config2task: Optional[Dict[str, str]] = None):
+    def __init__(self, config_class: luigi.Config, parameter_alias: Optional[Dict[str, str]] = None):
         """
         Decorates task to inherit parameter value of `config_class`.
 
         * config_class: Inherit parameter value of this task to decorated task. Only parameter values exist in both tasks are inherited.
-        * param_config2task: Dictionary to map paramter names between config_class and decorated task.
-                             key: config_class's parameter name. value: decorated task's parameter name.
+        * parameter_alias: Dictionary to map paramter names between config_class task and decorated task.
+                           key: config_class's parameter name. value: decorated task's parameter name.
         """
 
         self._config_class: luigi.Config = config_class
-        self._param_config2task: Dict[str, str] = param_config2task if param_config2task is not None else {}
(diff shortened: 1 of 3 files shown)


test/test_config_params.py
```diff
@@ -25,6 +25,12 @@ class Inherited(gokart.TaskOnKart):
     param_b = luigi.Parameter(default='overrided')
 
 
+@inherits_config_params(ConfigClass, param_config2task={'param_a': 'param_d'})
```
vaaaaanquish, line 28: When would you be happy with this feature?

vaaaaanquish (changes requested):
I'd like you to write a document :)

test/test_config_params.py
mski-iksm: @vaaaaanquish 
This would be useful when `ConfigClass` and Wrapped class has different parameter name but want to have common value.

gokart/config_params.py
```diff
@@ -1,23 +1,26 @@
...
+
 import luigi
 
 import gokart
 
 
 class inherits_config_params:
-    def __init__(self, config_class: luigi.Config):
-        self.config_class: luigi.Config = config_class
+    def __init__(self, config_class: luigi.Config, param_config2task: Optional[Dict[str, str]] = None):
```
Hi-king, line 9: @mski-iksm [nits] Plz write docstring about args here :)

test/test_config_params.py
mski-iksm: @vaaaaanquish 
I've added use case to the docs.
https://github.com/m3dev/gokart/pull/193/files#diff-bd31f137cd4066a176d3de6d9729841a660ed1416d3a26cf9bd0c820211ad8b8R57

vaaaaanquish (approved):
LGTM

gokart/config_params.py
mski-iksm: @Hi-king 
I've changed `param_config2task` to `parameter_alias`. Please review!

Hi-king (approved):
LGTM!

PR #194, opened 2021-04-14
🐛 Hotfix: Should use set() for empty set declaration

Hi-king:
- An one-line fix, `set()`
- small fixes on test
- CI runs on any PRs

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches: [ master ]
   pull_request:
-    branches: [ master ]
 
 jobs:
   tests:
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -4,7 +4,6 @@ on:
   push:
     branches: [ master ]
   pull_request:
-    branches: [ master ]
```
Hi-king, line 7: Changed here because CI should be run in any PRs 
@hirosassa

hirosassa: 👍

hirosassa (approved):
Thanks! LGTM

PR #205, opened 2021-05-02
Add options to check task definition to create an output

hiro-o918:
## What
Add a feature to create a hash of a task.
The hash is created by `cloudpickle.dumps` to the class object of a task.

## Why
Current implementation assumes that an output is determined by input parameters and requirements.
But this assumption does not work, when the implementation of a task changes.

As a results, some issues can occur.
- In local development, we need to remove output `.pkl` to rerun tasks
  - `modification_time_check` can resolve this in some aspects (I like this option)
- In production with some storage (e.g. S3), backfilling or re-running can be difficult because output `.pkl` must be removed as well as local development.

So, adding task definition to hash resolves these issues and re-reruns tasks safely.

## Discussion
- this feature depends on `cloudpickle`

### DIFF
--- a/gokart/task.py
+++ b/gokart/task.py
@@ -5,6 +5,7 @@ from importlib import import_module
 from logging import getLogger
 from typing import Any, Callable, Dict, List, Optional, Set, Union
 
+import cloudpickle
 import luigi
 import pandas as pd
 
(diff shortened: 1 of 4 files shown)


vaaaaanquish:
@hiro-o918 Thanks great suggestions. Is it possible to apply a yapf formatter?

vaaaaanquish:
LGTM. I'll wait for the @Hi-king review.

mski-iksm:
@hiro-o918 
LGTM! Thank you for your contribution!

hirosassa (approved):
Thank you for your great contribution!

Hi-king:
Cool solution! For workaround, currently we're using `__version=IntParameter(default=2)` for declaring the change of task definition 💦  This way looks more sophisticated to me !

hiro-o918:
Thank you for your fast your reviews! :rocket:

hirosassa:
@hiro-o918 [gokart 1.0.1](https://pypi.org/project/gokart/1.0.1/) including your great contribution is released. Thanks again!

Issue #211, opened 2021-06-12
Feature request: Set logging level for gokart.build


TaylerUva:
Allow setting of logging level in gokart.build. I would like to be able to see info logs I've created in my tasks without having to see gokart debug messages. This is useful for Juptyer Notebooks.

https://github.com/m3dev/gokart/blob/575207f5d348b8c1afbb6d1af3777c67230e15ed/gokart/build.py#L41-L51

vaaaaanquish:
@TaylerUva [ASK]
There is a way to make a config, what do you think of it?
example: https://gist.github.com/vaaaaanquish/8719a3470aa83c7985c5416ece0299c0

Or do you think `gokart.build` argument would be better?

TaylerUva:
For my use case a config would be better. As long as that will work with logging from other classes.

i.e. `I'm a log!` will get logged when `do_thing()` is called in `TaskOnKart.run`
```python
from logging import getLogger


class MyClass:
    logger = getLogger(__name__)
    def do_thing():
        self.logger.info("I'm a log!")
```

Doing it through a config will also make your existing documentation more true. (I had to look at source code to realize the logging config was ignored by build)

vaaaaanquish:
> Doing it through a config will also make your existing documentation more true. (I had to look at source code to realize the logging config was ignored by build)

That's right. I think so too, that's very confusing.

vaaaaanquish:
I was reading luigi. It looks like putting the logger configuration to `luigi.build` args.

PR #212, opened 2021-06-12
add loglevel settings

hirosassa:
fixes https://github.com/m3dev/gokart/issues/211

Thanks to @TaylerUva

### DIFF
--- a/docs/intro_to_gokart.rst
+++ b/docs/intro_to_gokart.rst
@@ -147,3 +147,13 @@ The :func:`~gokart.build` is inline code.
             self.dump(self.param)
 
     gokart.build(SampleTask(param='hello'), return_value=False)
+
+
+To output logs of each tasks, you can pass `~log_level` parameter to `~gokart.build` as follows:
+
+.. code:: python
+
+    gokart.build(SampleTask(param='hello'), return_value=False, log_level=logginge.DEBUG)
+
+
+This feature is very useful for running `~gokart` on jupyter notebook.
--- a/gokart/build.py
+++ b/gokart/build.py
@@ -9,14 +9,18 @@ from gokart.task import TaskOnKart
 from gokart.utils import check_config, read_environ
 
 
-class HideLogger:
-    def __init__(self, verbose: bool):
+class LoggerConfig:
+    def __init__(self, verbose: bool, level: Optional[int]):
         self.verbose = verbose
         self.logger = getLogger(__name__)
         self.default_level = self.logger.level
+        self.log_level_in_build = level
 
     def __enter__(self):
-        if not self.verbose:
+        if self.log_level_in_build is not None:
+            logging.disable(self.log_level_in_build)
+            self.logger.setLevel(self.log_level_in_build)
+        elif not self.verbose:
             logging.disable(sys.maxsize)
             self.logger.setLevel(logging.CRITICAL)
         return self
@@ -38,7 +42,7 @@ def _reset_register(keep={'gokart', 'luigi'}):
                                          if x.__module__.split('.')[0] in keep]  # avoid TaskClassAmbigiousException
 
 
-def build(task: TaskOnKart, verbose: bool = False, return_value: bool = True, reset_register: bool = True) -> Optional[Any]:
+def build(task: TaskOnKart, verbose: bool = False, return_value: bool = True, reset_register: bool = True, log_level: Optional[int] = None) -> Optional[Any]:
     """
     Run gokart task for local interpreter.
     """
@@ -46,6 +50,6 @@ def build(task: TaskOnKart, verbose: bool = False, return_value: bool = True, re
         _reset_register()
     read_environ()
     check_config()
-    with HideLogger(verbose):
+    with LoggerConfig(verbose, level=log_level):
         luigi.build([task], local_scheduler=True)
     return _get_output(task) if return_value else None
--- a/gokart/build.py
+++ b/gokart/build.py
@@ -16,11 +16,12 @@ class LoggerConfig:
         self.level = level
 
     def __enter__(self):
-        if self.level is not None:
+        if self.level is None:
+            logging.disable(sys.maxsize)
+            self.logger.setLevel(logging.CRITICAL)
+        else:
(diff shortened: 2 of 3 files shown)


gokart/build.py
```diff
@@ -9,14 +9,18 @@
...
         self.logger = getLogger(__name__)
         self.default_level = self.logger.level
+        self.log_level_in_build = level
 
     def __enter__(self):
-        if not self.verbose:
+        if self.log_level_in_build is not None:
+            logging.disable(self.log_level_in_build)
+            self.logger.setLevel(self.log_level_in_build)
+        elif not self.verbose:
```
hirosassa, line 23: [help wanted]
In this implementation, `log_level_in_build` is prioritized than `verbose`.
Which do you prefer this or `verbose` is prioritized one?

gokart/build.py
```diff
@@ -9,14 +9,18 @@
 from gokart.utils import check_config, read_environ
 
 
-class HideLogger:
-    def __init__(self, verbose: bool):
+class LoggerConfig:
+    def __init__(self, verbose: bool, level: Optional[int]):
```
vaaaaanquish, line 13: [NITS] Maybe not int?

gokart/build.py
vaaaaanquish: How do you think about remove `verbose`, and setting the default for `loglevel`?

```python
class LoggerConfig:
    def __init__(self, level: Optional[int]=logging.CRITICAL):
    ....

    def __enter__(self):
        if self.level==logging.CRITICAL:
            logging.disable(sys.maxsize)
        else:
            logging.disable(self.level)
        self.logger.setLevel(self.level)
        return self
```

test/test_build.py
```diff
@@ -30,5 +32,16 @@ def test_build(self):
...
 
+class LoggerConfigTest(unittest.TestCase):
+    def test_logger_config(self):
+        verbose = False
+        level = None
+        for verbose, level, expected in ((False, None, logging.CRITICAL), (True, None, logging.NOTSET), (False, logging.DEBUG, logging.DEBUG),
+                                         (True, logging.DEBUG, logging.DEBUG)):
+            with self.subTest(verbose=verbose, level=level, expected=expected):
+                with LoggerConfig(verbose, level) as lc:
+                    self.assertEqual(lc.logger.level, expected)
```
vaaaaanquish, line 43: Thanks add unit test. 🆒

gokart/build.py
vaaaaanquish: Backward compatibility will be lost, though. 
If you don't want to do that, I agree with @hirosassa 's  implementation.

vaaaaanquish:
Adding more interfaces to parts that are already configurable is likely to lead to bugs.  
We need to think about whether `gokart.build` should be in charge of this.
https://github.com/m3dev/gokart/issues/211#issuecomment-860014657

TaylerUva:
Any updates on this?

gokart/build.py
Hi-king: I'm standing with @vaaaaanquish 's suggestion. Maintaining both verbose and level options are confusing to me.

Hi-king: I'm standing with @vaaaaanquish 's suggestion. Maintaining both verbose and level options are confusing to me.

Hi-king (changes requested):

gokart/build.py
hirosassa: @vaaaaanquish @Hi-king Thanks for your fruitful FBs. fixed!

gokart/build.py
```diff
@@ -9,16 +9,18 @@
...
         self.default_level = self.logger.level
+        self.level = level
 
     def __enter__(self):
-        if not self.verbose:
+        if self.level is not None:
+            logging.disable(self.level)
+            self.logger.setLevel(self.level)
+        if self.level == logging.CRITICAL:
             logging.disable(sys.maxsize)
```
TaylerUva, line 23: Why disable if set to critical?

Wouldn't the following be better?
```suggestion
        if self.level is None:
            logging.disable(sys.maxsize)
```

TaylerUva (changes requested):

gokart/build.py
hirosassa: Thanks! Your suggestion is right! Fixed!

hirosassa:
@vaaaaanquish @Hi-king plz review!

gokart/build.py
```diff
@@ -9,16 +9,19 @@
 from gokart.utils import check_config, read_environ
 
 
-class HideLogger:
-    def __init__(self, verbose: bool):
-        self.verbose = verbose
+class LoggerConfig:
+    def __init__(self, level: Optional[int]):
```
Hi-king, line 13: @hirosassa How about     `def __init__(self, level: int = logging.CRITICAL): ` instead of Optional ?

Hi-king: Are there any special reason to allow None instead of CRITICAL?

hirosassa: There's no specific reason to allow that. Fixed in  60f7a65.

vaaaaanquish (approved):
LGTM. thx

vaaaaanquish:
@hirosassa plz write documents!

docs/intro_to_gokart.rst
```diff
@@ -147,3 +147,13 @@ The :func:`~gokart.build` is inline code.
...
+
+
+To output logs of each tasks, you can pass `~log_level` parameter to `~gokart.build` as follows:
+
+.. code:: python
+
+    gokart.build(SampleTask(param='hello'), return_value=False, log_level=logginge.DEBUG)
+
+
+This feature is very useful for running `~gokart` on jupyter notebook.
```
hirosassa, line 159: @vaaaaanquish docs added!

hirosassa:
@Hi-king simplified in 12cd1d6. plz check this

Hi-king (approved):

Hi-king:
@hirosassa Thx for continuous updates
@TaylerUva @vaaaaanquish Thx for reviews

Merged :)

PR #217, opened 2021-06-17
Enable pandas 1.x.x

davegrays:


hirosassa:
@davegrays Thanks for your contribution! Please run following command and commit `poetry.lock` file.

```bash
poetry lock
```

pyproject.toml
```diff
@@ -19,7 +19,7 @@ python = "^3.7"
 luigi = "*"
 boto3 = "*"
 slack-sdk = "^3.0"
-pandas = "<1.2"
+pandas = "^1.0"
```
hirosassa, line 22: Since `pandas` needs Python >=3.7.1, you should specify python version as `python = "^3.7.1"`

https://pypi.org/project/pandas/

hirosassa:
@davegrays This is friendly reminder. In my investigation, this change will cause several test failures. 
Do you want to fix these? or I can take this PR over.

hirosassa:
@davegrays I'll close this PR because this PR is not active. Anyway thanks for your contribution!

PR #218, opened 2021-06-21
Fix add_config behavior

e-mon:
The following error occurs when adding a configuration file with environment variables.

base.ini
```ini
[Example]
param = %(param)s
```

example.py
```pyuthon
import gokart

class Example(gokart.TaskOnKart):
    param = luigi.Parameter()

    def run(self):
        self.dump('Hello, world!')

gokart.add_config('base.ini')
task = Example()
output = gokart.build(task)
```

erorr
```
configparser.InterpolationMissingOptionError: Bad value substitution: option 'param' in section 'Example' contains an interpolation key 'test_param' which is not a valid option name. Raw value: '%(test_param)s'
```

I think this is because the environment variables are not loaded before the parameter parsing, so I fixed `config_add` function.

Please review.

### DIFF
--- a/gokart/utils.py
+++ b/gokart/utils.py
@@ -24,3 +24,4 @@ def add_config(file_path: str):
     _, ext = os.path.splitext(file_path)
     luigi.configuration.core.parser = ext
     assert luigi.configuration.add_config_path(file_path)
+    read_environ()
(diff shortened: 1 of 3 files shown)


vaaaaanquish:
いいよ

vaaaaanquish (approved):

hirosassa:
Looks good!

hirosassa (approved):

hirosassa:
@e-mon Thanks for your contribution! merged!

PR #221, opened 2021-07-02
Fix build output error when using dicts

### LINKED ISSUE #220: [Bug] Failure when def output returns dict and using build
An attribute error is raised when creating a def output() that returns a dictionary and the task is run with gokart build.

```python
gokart/gokart/build.py", line 35, in _get_output
    return output.load()
AttributeError: 'dict' object has no attribute 'load'
```

```python
def _get_output(task: TaskOnKart) -> Any:
    output = task.output()
    if type(output) == list:
        return [x.load() for x in output]
    return output.load()
```
_get_output only checks for a list but output should also support tuple and dict

TaylerUva:
fixes #220

### DIFF
--- a/gokart/build.py
+++ b/gokart/build.py
@@ -28,8 +28,10 @@ class HideLogger:
 
 def _get_output(task: TaskOnKart) -> Any:
     output = task.output()
-    if type(output) == list:
-        return [x.load() for x in output]
+    if isinstance(output, list) or isinstance(output, tuple):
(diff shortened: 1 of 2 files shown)


hirosassa (approved):
@TaylerUva Thanks for your contribution! Looks great!

vaaaaanquish:
LGTM.

vaaaaanquish (approved):

PR #223, opened 2021-07-11
add indexedfeather file processor

mski-iksm:
[What?]
I've added indexed feather file processor which is capable to store index data.

[Why?]
The existing FeatherFileProcessor cannot save index information since the index must be 0,1,2,.... when saving as feather.

Please review.
@Hi-king @hirosassa @vaaaaanquish @e-mon

### DIFF
--- a/docs/task_on_kart.rst
+++ b/docs/task_on_kart.rst
@@ -100,6 +100,10 @@ Please refer to :func:`~gokart.task.TaskOnKart.make_target` and described later 
     By default, file path is inferred from "__name__" of the script, so ``output`` method can be omitted.
     Please refer to :doc:`tutorial` section.
 
+.. note::
+    When using `.feather`, index will be converted to column at saving and restored to index at loading.
+    If you don't prefere saving index, set `store_index_in_feather=False` parameter at `gokart.target.make_target()`.
+
 
 TaskOnKart.load
 ----------------
--- a/docs/task_on_kart.rst
+++ b/docs/task_on_kart.rst
@@ -85,6 +85,13 @@ It is also possible to specify a file format other than pkl. The supported file 
 - .gz
 - .json
 - .xml
+- .npz
+- .parquet
+- .feather: This file type cannot save index of pd.DataFrame. Use `.indexedfeather` instead when index must be unchanged.
+- .indexedfeather: feather file format with index.
+- .png
+- .jpg
+
 
 If dump something other than the above, can use :func:`~gokart.TaskOnKart.make_model_target`.
 Please refer to :func:`~gokart.task.TaskOnKart.make_target` and described later Advanced Features section.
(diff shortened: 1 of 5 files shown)


hirosassa:
Looks good!

hirosassa (approved):

vaaaaanquish:
Is there any reason why reset_index should not be used?

I think the code is good, but there is no additional documentation and I don't know the usage scenarios.

vaaaaanquish (changes requested):
Add documents and ut

mski-iksm:
@vaaaaanquish 

> Is there any reason why reset_index should not be used?
> 
> I think the code is good, but there is no additional documentation and I don't know the usage scenarios.

If you have feather file which is dumped with FeatherFileProcessor and index processed with reset_index(), there is no way to know which column was index originally.
Using IndexedFeatherFileProcessor
- when dump(): index will converted to a column with certain prefix on the name
- when load(): a column with certain prefix on the name will be restored as index

By this behavior, you can use IndexedFeatherFileProcessor as like other FileProcessor without considering about index problem.

vaaaaanquish:
@mski-iksm I know how it works. But I can't think of a situation where more than two Tasks need `index` . Since indexing of pandas is not stable (implicit reset, pandas version,...and more), it is better to use column as the basic method for merging, grouping, ...


There are two reasons why I am cautious about this approach
- It's trying to force implementation of features that are sufficient for column.
- Adding the original extension will confuse the user > `.indexedfeather`

vaaaaanquish:
For example, can we consider extending the existing FeatherFileProcessor?

mski-iksm:
@vaaaaanquish 
>For example, can we consider extending the existing FeatherFileProcessor?

Yes, I will
- move index processing codes to FeatherFileProcessor, and remove IndexedFeatherFileProcessor
- add option to FeatherFileProcessor whether to process the index or not
- add note of FeatherFileProcessor about the index processing option to docs

mski-iksm:
@Hi-king @vaaaaanquish 
I've added test and docs. Please review!

docs/task_on_kart.rst
```diff
@@ -94,6 +100,10 @@ Please refer to :func:`~gokart.task.TaskOnKart.make_target` and described later
     By default, file path is inferred from "__name__" of the script, so ``output`` method can be omitted.
     Please refer to :doc:`tutorial` section.
 
+.. note::
+    When using `.feather`, index will be converted to column at saving and restored to index at loading.
+    If you don't prefere saving index, set `store_index_in_feather=False` parameter at `gokart.target.make_target()`.
```
vaaaaanquish, line 105: 🆒

vaaaaanquish (approved):

docs/task_on_kart.rst
hirosassa: 👍 +1:

mski-iksm:
@hirosassa @vaaaaanquish Thank you for reviewing! I'll merge this PR.

Issue #231, opened 2021-08-08
Proposing new feature: dump dependent Tasks information as a table


mski-iksm:
I thought dumping a dependent tasks information table after running tasks is convenient for post analysis.

Task cache information table is a pandas dataframe which has information such as task name, unique id, cache file path, parameter and processing time of completed tasks.
Implementation was considered in https://github.com/m3dev/gokart/pull/176, however the PR was closed due to difficulty in behavior setting.

mski-iksm:
I'll first implement this feature in `gokart.build()`.
Implementation to `gokart.run()` will be done in different issue/PR, since `gokart.run()` has difficulty in parameter settings.

PR #232, opened 2021-08-09
add new feature: dump dependent Tasks information as a table

mski-iksm:
I've added a new feature for dumping task cache information file at `gokart.build()`.

Task cache information file is a pandas dataframe which has information such as task name, unique id, cache file path, parameter and processing time of completed tasks.

issue: https://github.com/m3dev/gokart/issues/231


mski-iksm:
@Hi-king @vaaaaanquish @e-mon @hirosassa Please review.

docs/task_information.rst
```diff
@@ -170,3 +170,19 @@ the output could be like:
 Delete Unnecessary Output Files
 --------------------------------
 To delete output files which are not necessary to run a task, add option ``--delete-unnecessary-output-files``. This option is supported only when a task outputs files in local storage not S3 for now.
+
+
+
+Dump TaskInfo Table
+--------------------
+When using ``gokart.build()``, you can dump information of dependent tasks to a file as TaskInfo table.
```
hirosassa, line 178: [ask] Is it difficult to dump the information on standard out or error as a log, too?

vaaaaanquish:
Since these information can be recovered from log that are already there, looks unnecessary feature.
Please write usage scenarios in document.

gokart/task_info.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def _make_task_info_table(task: TaskOnKart, ignore_task_names: Optional[List[str]]):
+    task_info_list = _make_task_info_list(task, set(), ignore_task_names=ignore_task_names)
+    return pd.DataFrame(task_info_list)
+
+
+def dump_task_info_table(task: TaskOnKart, task_info_dump_path: Optional[str], task_info_ignore_task_names: Optional[List[str]]):
+    if task_info_dump_path is None:
+        return None
```
vaaaaanquish, line 51: It's better to make the decision outside of function.

docs/task_information.rst
```diff
@@ -170,3 +170,19 @@ the output could be like:
...
+TaskInfo table is a pandas DataFrame format table containing `task name`, `cache unique id`, `cache file path`, `task parameters`, and `task processing time`.
+
+To dump TaskInfo table, set ``task_info_dump_path`` parameter at ``gokart.build()``.
+
+- path destination can be `local`, `S3`, or `GCS`
+- file extension can be any type that gokart file processor accepts, including `csv`, `pickle`, or `txt`.
+
+.. code:: python
+
+    gokart.build(task=task, task_info_dump_path="./task_info.csv", task_info_ignore_task_names=None)
```
vaaaaanquish, line 188: Write a description of `task_info_ignore_task_names` to here.

docs/task_information.rst
```diff
@@ -170,3 +170,19 @@ the output could be like:
...
+
+Dump TaskInfo Table
+--------------------
+When using ``gokart.build()``, you can dump information of dependent tasks to a file as TaskInfo table.
+TaskInfo table is a pandas DataFrame format table containing `task name`, `cache unique id`, `cache file path`, `task parameters`, and `task processing time`.
+
+To dump TaskInfo table, set ``task_info_dump_path`` parameter at ``gokart.build()``.
+
+- path destination can be `local`, `S3`, or `GCS`
+- file extension can be any type that gokart file processor accepts, including `csv`, `pickle`, or `txt`.
```
vaaaaanquish, line 184: Please add a link to the file proceccer documentation.

vaaaaanquish (changes requested):

docs/task_information.rst
vaaaaanquish: for example: https://gokart.readthedocs.io/en/latest/task_on_kart.html#taskonkart-make-target

mski-iksm:
@vaaaaanquish 

> Since these information can be recovered from log that are already there, looks unnecessary feature.

I think you are talking about using `gokart.make_tree_info()` which is very close feature.
However,
- it's not formatted and difficult to read
- In some case, getting the root task with all parameters set is difficult to reproduce

so, I think this feature is needed for productivity.


I do agree with `_make_task_info_list()` duplicated with  `gokart.make_tree_info()`, so I think replacing backend with `gokart.make_tree_info()`. How do you think?

vaaaaanquish:
@mski-iksm 
> I do agree with _make_task_info_list() duplicated with gokart.make_tree_info(), so I think replacing backend with gokart.make_tree_info()

This is very cool. If `make_tree_info` has the ability to return `dataframes`, It's a big implement.

mski-iksm:
@vaaaaanquish 
I've modified `dump_task_info_table` to use the same backend as `make_tree_info`.
`dump_task_info_table` is mostly same with `make_tree_info`, except it dumps the tree_info as pandas.DataFrame.

During this process, I've refactored `make_tree_info` and fixed some tests.

docs/task_information.rst
mski-iksm: @hirosassa 
If you want to dump the information as a standard out, I think of using `make_tree_info` may be better.

mski-iksm:
I'll close this PR and separate it into following 2 PRs.
- Refactor `make_tree_info` + move to different package
- implement feature of dumping task info table as a option of `make_tree_info`

PR #233, opened 2021-08-09
fix LoggerConfig inconsistency bug

mski-iksm:
`gokart.build()` has a feature to disable low important logs.
However, the implementation has inconsistency, 
- logging.disable(self.level): disable any logs with level same or below the self.level
- self.logger.setLevel(self.level): enable any logs same or above the self.level

This is because logging.Logger.setLevel and logging.disable has the following difference.

> setLevel: Logging messages which **_are less severe than_** level will be ignored
> logging.disable: disable all logging calls of **_severity level and below_**

This PR fix the inconsistency by `disabling any logs with level below the self.level`

### DIFF
--- a/gokart/build.py
+++ b/gokart/build.py
@@ -16,12 +16,12 @@ class LoggerConfig:
         self.level = level
 
     def __enter__(self):
-        logging.disable(self.level)
+        logging.disable(self.level - 10)
         self.logger.setLevel(self.level)
         return self
 
     def __exit__(self, exception_type, exception_value, traceback):
-        logging.disable(self.default_level)
+        logging.disable(self.default_level - 10)
         self.logger.setLevel(self.default_level)
 
 
--- a/gokart/build.py
+++ b/gokart/build.py
@@ -16,12 +16,12 @@ class LoggerConfig:
         self.level = level
 
     def __enter__(self):
-        logging.disable(self.level - 10)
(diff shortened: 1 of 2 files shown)


gokart/build.py
```diff
@@ -16,12 +16,12 @@ def __init__(self, level: int = logging.CRITICAL):
...
 
     def __enter__(self):
-        logging.disable(self.level)
+        logging.disable(self.level - 10)
         self.logger.setLevel(self.level)
         return self
 
     def __exit__(self, exception_type, exception_value, traceback):
-        logging.disable(self.default_level)
+        logging.disable(self.default_level - 10)
```
vaaaaanquish, line 24: Does it make sense to be 10? If not, use `sys.maxsize` to make the meaning clearer.

mski-iksm: @vaaaaanquish 
I think it should be 10,  because we want to change "disabling logs with level self.default_level and above" to "disabling logs with levels above self.default_level"

vaaaaanquish: I see. It seems like a good idea to put it in the comments.

vaaaaanquish (approved):
LGTM 🆒

hirosassa:
Thanks! LGTM

hirosassa (approved):

PR #250, opened 2021-09-14
Add requires to TaskInfo

mski-iksm:
I've added `requires` attribute to `TaskInfo`, so that information of `task.requires()` will be dumped to tree_info_table.

This is useful when a task requires 2 same tasks but with different parameter.
With key as index, you can get the unique_id of the two tasks separately.

```python
class _TaskInfoExampleTaskA(gokart.TaskOnKart):
    def requires(self):
        return dict(task1=_TaskInfoExampleTaskA(param=1), task2=_TaskInfoExampleTaskA(param=2))
```

In the above example, the requires column of tree_info_table will have dictionary with same keys as return() (i.e. task1, task2) and object of name and unique_id (i.e. RequiredTask(name, unique_id) object) as values.
In this way, we can distinguish between two tasks by key and get the unique_id.

Please review.

### DIFF
--- a/gokart/tree/task_info_formatter.py
+++ b/gokart/tree/task_info_formatter.py
@@ -1,6 +1,6 @@
 import warnings
 from dataclasses import dataclass
-from typing import List, Optional, Set
+from typing import Dict, List, NamedTuple, Optional, Set, Union
 
 import luigi
 import pandas as pd
@@ -18,6 +18,7 @@ class TaskInfo:
     processing_time: str
     is_complete: str
     task_log: dict
+    requires: Union['RequiredTask', List['RequiredTask'], Dict[str, 'RequiredTask']]
     children_task_infos: List['TaskInfo']
 
     def get_task_id(self):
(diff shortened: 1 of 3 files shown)


gokart/tree/task_info_formatter.py
```diff
@@ -36,7 +37,26 @@ def task_info_dict(self):
...
+    if isinstance(requires, TaskOnKart):
+        return RequiredTask(name=requires.__class__.__name__, unique_id=requires.make_unique_id())
+
+    if isinstance(requires, list) or isinstance(requires, tuple):
+        return [_make_requires_info(requires=item) for item in requires]
+
+    if isinstance(requires, dict):
+        return {key: _make_requires_info(requires=item) for key, item in requires.items()}
+
+    raise TypeError(f'`requires` has unexpected type {type(requires)}. Must be `TaskOnKart`, `List[TaskOnKart]`, or `Dict[str, TaskOnKart]`')
```
hirosassa, line 59: IMO, raising TypeError case test is needed

hirosassa (changes requested):
Thank you for the contribution! Commented.

gokart/tree/task_info_formatter.py
mski-iksm: I've added test_task_info_formatter.py for testing _make_requires_info() !

Hi-king (approved):
LGTM

gokart/tree/task_info_formatter.py
hirosassa: Thanks! Confirmed

hirosassa (approved):
Looks Great!

Hi-king:
@mski-iksm Merged. Thx!

Issue #216, opened 2021-06-17
Feature Request: Tasks hash ids update when run function definition changes


TaylerUva:
When the code signature of the run function changes in a task, this should update the hash value since the hash is different than its last run.

```python
class TaskA(gokart.TaskOnKart):

    param = luigi.IntParameter()

    def run():
        sum = self. param + 3
        self.dump(sum)
```

```python
class TaskA(gokart.TaskOnKart):

    param = luigi.IntParameter()

    def run():
        sum = self. param + 2
        self.dump(sum)
```

These should have different hash ids because the internal logic has changed.

vaaaaanquish:
It's difficult issue to solve because of It has side effects.

With the following code, the user will have to think about whether the hash value will change when the version of numpy changes.
```
import numpy

class TaskA(gokart.TaskOnKart):

    param = luigi.IntParameter()

    def run():
        self.dump(np.ndarray([1]))
```
We also have to choose how much of the internal changes of class and decorator you want to include in the hash. This is because gokart is used in a variety of settings, including experiments, competitions, productions and more.

How far do you think we should look?

PR #252, opened 2021-10-13
Tasks hash ids update when run function definition changes

ujiuji1259:
I’ve added `get_own_code` method so that unique id can be updated when the task definition changes (requested in #216 ).

In the following example (in `examples/gokart_notebook_example.ipynb`),  `get_own_code` got the codes of `run`, `requires`.
By adding this into `dependencies` of hash_id, the task rerun without manually changing `__version` when the script is changed.
スクリーンショット 2021-10-13 8 30 46

This PR is related to #205 from @hiro-o918

### DIFF
--- a/gokart/task.py
+++ b/gokart/task.py
@@ -275,7 +275,7 @@ class TaskOnKart(luigi.Task):
         def has_sourcecode(obj):
             return inspect.ismethod(obj) or inspect.isfunction(obj) or inspect.isframe(obj) or inspect.iscode(obj)
 
-        return set([inspect.getsource(t) for _, t in inspect.getmembers(target_class, has_sourcecode)])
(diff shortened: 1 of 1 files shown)


hirosassa:
@ujiuji1259 Thank you for the GREAT contribution! 
Please add unit test for the `get_code` and `make_unique_id` to confirm this feature runs correctly.

mski-iksm:
Thank you for great contribution!

I think the description should be something like, 

> If this is true, even if all outputs are present, this task will be executed if any changes have been made to the code.

instead of

> If this is true, this task will not run only if all input and output files exist, and this task class is modified.

Hi-king:
@ujiuji1259 Thx for great feature! We'll release new version including this feature :)

hiro-o918:
Thank you for awesome new features!
I’m looking forward to using it!☺️

PR #259, opened 2021-11-06
Passing additional arguments with gokart.build() (like workers=...)

### LINKED ISSUE #251: How to run tasks in parallel
Let's assume the following example:

```python
import gokart
import luigi
import time

class DemoTask(gokart.TaskOnKart):
  name = luigi.Parameter()

  def run(self,):
    print(f"demo running for {self.name}")
    time.sleep(1)
    self.dump(f"Demo -> {self.name}")


class Prepare(gokart.TaskOnKart):
  names = list(range(0,20,1))

  def requires(self):
    return {f"task_{n}": DemoTask(name=str(n)) for n in self.names}

  def run(self):
    self.dump('DONE')


prep = Prepare()
output = gokart.build(task=prep)
print(output)
```

All `DemoTasks` are executed randomly but one after another. Is there a way to execute them in parallel?

Hi-king:
Resolves #251 

With this fix, we can pass all the parameters which luigi.build can handle :)

### DIFF
--- a/test/test_build.py
+++ b/test/test_build.py
@@ -41,6 +41,14 @@ class _DummyFailedTask(gokart.TaskOnKart):
         raise RuntimeError
 
 
+class _ParallelRunner(gokart.TaskOnKart):
+    def requires(self):
+        return [_DummyTask(param=str(i)) for i in range(10)]
+
+    def run(self):
+        self.dump('done')
+
+
 class RunTest(unittest.TestCase):
     def setUp(self):
         luigi.configuration.LuigiConfigParser._instance = None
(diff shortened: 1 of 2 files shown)


test/test_build.py
```diff
@@ -57,6 +65,10 @@ def test_build(self):
         output = gokart.build(_DummyTask(param=text), reset_register=False)
         self.assertEqual(output, text)
 
+    def test_build_parallel(self):
+        output = gokart.build(_ParallelRunner(), reset_register=False, workers=20)
+        self.assertEqual(output, 'done')
+
```
hirosassa, line 71: nice test!

hirosassa (approved):
Looks good!

mski-iksm (approved):
Looks great!

hirosassa:
@Hi-king Merged!

PR #262, opened 2021-12-29
add vscode and ds_store to gitignore

mski-iksm:
I've added vscode setting folder and .DS_Store to git_ignore.

Please review.


.gitignore
```diff
@@ -114,4 +114,10 @@ examples/resources
...
 # temporary data
-temporary.zip
\ No newline at end of file
+temporary.zip
+
+# vscode
+.vscode/
+
+# DS_Store
+.DS_Store
```
hirosassa, line 123: @mski-iksm This kind of ignore entries should be set in personal git configurations (not in specific repo's config).

hirosassa: ex) personal environment specific configurations
https://github.com/github/gitignore/tree/main/Global

IMO, this kind of ignore entries put in `$HOME/.config/git/ignore`
ref: https://git-scm.com/docs/gitignore

mski-iksm: @hirosassa 
Thank you for the comment. I see how it should be managed.

I will close this PR.

PR #270, opened 2022-01-30
Add exist check before removing file

mski-iksm:
I've added existence check to `ModelTarget._remove()` as like in `SingleFileTarget._remove()`.

Also, I've changed existence check of `SingleFileTarget._remove()` to use `self._exists()`.

Please review!

### DIFF
--- a/gokart/target.py
+++ b/gokart/target.py
@@ -102,7 +102,7 @@ class SingleFileTarget(TargetOnKart):
             self._processor.dump(obj, f)
 
     def _remove(self) -> None:
-        if self._target.exists():
+        if self._exists():
             self._target.remove()
 
     def _last_modification_time(self) -> datetime:
@@ -149,7 +149,8 @@ class ModelTarget(TargetOnKart):
         self._remove_temporary_directory()
 
     def _remove(self) -> None:
-        self._zip_client.remove()
+        if self._exists():
+            self._zip_client.remove()
 
     def _last_modification_time(self) -> datetime:
         return _get_last_modification_time(self._zip_client.path)


hirosassa:
LGTM!

hirosassa (approved):

gokart/target.py
```diff
@@ -102,7 +102,7 @@ def _dump(self, obj) -> None:
             self._processor.dump(obj, f)
 
     def _remove(self) -> None:
-        if self._target.exists():
+        if self._exists():
```
Hi-king, line 105: @mski-iksm how about

```suggestion
        if self.exists():
```

(for all other occurrences)

because _exists() is inner method, exists() is better, I think.

Hi-king (changes requested):

gokart/target.py
hirosassa: @mski-iksm ping?

mski-iksm: @Hi-king I thought self.exists() is better because it closes the relationship within this class without creating relationships with the super class. How do you think?

Hi-king: @mski-iksm OK. I just concern about future implementation between exist() and _exist() will be skipped. But might be too early concern.

gokart/target.py
```diff
@@ -149,7 +149,8 @@ def _dump(self, obj) -> None:
         self._remove_temporary_directory()
 
     def _remove(self) -> None:
-        self._zip_client.remove()
+        if self._exists():
```
Hi-king, line 152: @mski-iksm [ASK] Is it OK not to warn or raise Exception? This looks like not intended action to remove not existing  files.

gokart/target.py
mski-iksm: @Hi-king I understood your concern  and also think that every class that inherit TargetOnKart have to do existence check is troublesome. I will move the existence check of _remove() to remove(). Thank you for the comment!

gokart/target.py
mski-iksm: @Hi-king I think raising Exception is not good idea because this remove method is always called when TaskOnKart.rerun parameter is True.
https://github.com/m3dev/gokart/blob/master/gokart/task.py#L108

If we are to raise exception, we need to add existence check at TaskOnKart before calling target.remove().
But I think there is no good motivation of changing the attitude of TargetOnKart and TaskOnKart.
How do you think of this?

Hi-king: @mski-iksm 

* Existence check on rerun seems to be good idea for me because it will be more clear what `rerun` intended to.
* As the same manner as standard library, Introduce flag `remove(missing_ok=True)` can be a solution
   * Pathlib has that interface https://docs.python.org/ja/3/library/pathlib.html#pathlib.Path.unlink
   * os.unlink raises FileNotFoundError
* It's OK to go with current solution :)

Hi-king: @mski-iksm OK. Thank you for clarifying the point!

Hi-king (approved):

PR #271, opened 2022-01-30
Bugfix: assert when redis_timeout is too short

mski-iksm:
`redis_timeout` which defines the time before redis lock evaporates, cannot be shorter than 10 seconds, because redis lock is extended every 10 seconds. https://github.com/m3dev/gokart/blob/master/gokart/redis_lock.py#L58

This PR adds an assert to check that the timeout is not too short.

Please review!


hirosassa (approved):
Thanks for the fixing! LGTM!

Hi-king:
@mski-iksm cool! thx

@hirosassa thx for reviews

Hi-king (approved):

Issue #265, opened 2022-01-14
The redis cache lock reduces task parallelism and causes unnecessary latency in execution


mski-iksm:
## Problem

Using task cache locking prevents cache conflicts even when multiple gokart tasks are running in parallel. However, if multiple tasks running in parallel depend on a same task, read process may cause congestion, and parallelism may be compromised.

The figure below shows a case where the problem occurs.
Suppose app A, B and C are running on different nodes and are trying to do a read() of the same task. Although the three apps are running in parallel, the task lock causes congestion.

![image](https://user-images.githubusercontent.com/34547057/149538790-787d744a-311a-475f-91c8-5797dd59b16e.png)

## How to fix the problem

Use different procedures to get locks when writing, reading, and deleting the cache as illustrated in the figures below.

![image](https://user-images.githubusercontent.com/34547057/149538896-001debdf-7f10-4b4d-a6ed-8196627ee805.png)

_remove() is the method to remove the cache which occurs when rerun parameter is set to True in the task. 

As mentioned in case 4 and 5 below, rerun is prone to cause problems when combined with the cache lock feature. Therefore, we would like to have a warning when rerun is set when the cache redis lock function is turned on.

### case1: When multiple writing tasks happen at the same time

Suppose, app A and B are both writing tasks.

In such a case, one of the tasks will wait until the other completes its writing process to prevent cache collisions.

![image](https://user-images.githubusercontent.com/34547057/149538959-db2ea450-9ff1-49cd-983d-ac4f5299563f.png)

### case2: When a writing task and reading tasks happen

Suppose, app A is a writing task and apps B and C are the reading tasks. Consider the case where task B and C try to read the cache at almost the same time while task A is executing a write operation.

In this case, tasks B and C will wait for task A to complete writing. This is because before _load(), there is a process that takes lock for a moment and releases it immediately.
Since tasks B and C do not hold the lock the whole time while they are reading, but release it immediately, they can start the other _load() without waiting for the completion of either _load(). This is useful when _load() takes a long time.

![image](https://user-images.githubusercontent.com/34547057/149538991-9ad6666c-3831-4a15-b4d2-3988c3c78c56.png)

### case3: When a writing task happens after a reading task

Contrary to the previous case, consider the case where app B for writing occurs while app A for reading is running first.

[...]

PR #276, opened 2022-02-23
Feature/implement fast load lock

mski-iksm:
issue: https://github.com/m3dev/gokart/issues/265

I've implemented redis lock following above issue.
Please review!

### DIFF
--- a/gokart/redis_lock.py
+++ b/gokart/redis_lock.py
@@ -1,6 +1,7 @@
+import functools
 import os
 from logging import getLogger
-from typing import NamedTuple
+from typing import Callable, NamedTuple
 
 import redis
 from apscheduler.schedulers.background import BackgroundScheduler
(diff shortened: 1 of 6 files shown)


gokart/redis_lock.py
```diff
@@ -80,6 +92,70 @@ def extend_lock():
     return wrapper
 
 
+def wrap_with_run_lock(func, redis_params: RedisParams):
+    return _wrap_with_lock(func=func, redis_params=redis_params)
+
+
+def wrap_with_dump_lock(func: Callable, redis_params: RedisParams, exist_check: Callable):
+    # When TargetOnKart.dump() is called, dump() must be wrapped with redis lock and cache existance check.
+    # https://github.com/m3dev/gokart/issues/265
```
hirosassa, line 101: This comment and detailed description of this feature should documented on `docstring` and gokart doc.

gokart/redis_lock.py
```diff
@@ -80,6 +92,70 @@ def extend_lock():
...
+            redis_lock.release()
+            scheduler.shutdown()
+            raise e
+
+    return wrapper
+
+
+def wrap_with_remove_lock(func, redis_params: RedisParams):
+    # When TargetOnKart.remove() is called, remove() must be simply wrapped with redis lock.
+    # https://github.com/m3dev/gokart/issues/265
```
hirosassa, line 155: This should be documented, too.

hirosassa (changes requested):
@mski-iksm Thanks for great contribution!

I put a few comments for documentation. 
Could you also add doc here?
https://github.com/m3dev/gokart/blob/master/docs/using_task_cache_collision_lock.rst

gokart/redis_lock.py
mski-iksm: @hirosassa Thank you for the comment! I've moved these comments to the docstring.

These methods are only used internally and not to be used by users. So, I think it should not be included in the documents to make the documents simpler. How do you think of this?

hirosassa: @mski-iksm 
> I think it should not be included in the documents

OK! I understood 👍

hirosassa (approved):
LGTM!

gokart/redis_lock.py
```diff
@@ -80,6 +92,80 @@ def extend_lock():
...
+        scheduler = _set_lock_scheduler(redis_lock=redis_lock, redis_params=redis_params)
+
+        try:
+            logger.debug(f'Task lock of {redis_params.redis_key} locked.')
+            redis_lock.release()
+            logger.debug(f'Task lock of {redis_params.redis_key} released.')
+            scheduler.shutdown()
+            result = func(*args, **kwargs)
+            return result
+        except BaseException as e:
```
kitagry, line 152: Where do you think BaseException will occur from?

* If `redis_lock.release()` or `scheduler.shutdown()` in `try` block raise Exception, the same function in `except` block will occur again and overwrap the exception which might be other exception.
* If `func(*args, **kwargs)` raise Exception, it doesn't need to unlock because unlock method already called.

So, I think try block is unnecessary in this function.

gokart/redis_lock.py
```diff
@@ -80,6 +92,80 @@ def extend_lock():
...
+        try:
+            logger.debug(f'Task lock of {redis_params.redis_key} locked.')
+            if not exist_check():
+                func(*args, **kwargs)
+            redis_lock.release()
+            logger.debug(f'Task lock of {redis_params.redis_key} released.')
+            scheduler.shutdown()
+        except BaseException as e:
+            logger.debug(f'Task lock of {redis_params.redis_key} released with BaseException.')
+            redis_lock.release()
```
kitagry, line 125: I think it is better to call unlock function with finally block.

```python
try:
  logger.debug(...)
  if not exist_check():
    func(*args, **kwargs)
finally:
  redis_lock.release()
  logger.debug(...)
  scheduler.shutdown()
```

gokart/redis_lock.py
mski-iksm: @kitagry Thank you for checking carefully! 

This try block was for caching Exception in func(). Therefore, as you mentioned, this try block seems to be unnecessary. I will delete this try block!

gokart/redis_lock.py
mski-iksm: @kitagry I agree with your suggestion! I will change `_wrap_with_lock` also.

kitagry (approved):
LGTM

Hi-king:
@mski-iksm thx for cool improvement :)

@hirosassa @kitagry thx for reviews!

Merged

PR #289, opened 2022-05-16
fix: annotation of inherits_config_params

yuta100101:
`inherits_config_params` is not called with `TaskOnKart` **instance** but `TaskOnKart` **class itself** or its subclass.
This suppresses mypy error.
And `task_class` is better for the parameter name than `task`.
I think this change does not have a significant impact on usage because `inherits_config_params` is used as decorator.

### DIFF
--- a/gokart/config_params.py
+++ b/gokart/config_params.py
@@ -1,4 +1,4 @@
-from typing import Dict, Optional
+from typing import Dict, Optional, Type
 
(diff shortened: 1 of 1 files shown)


Hi-king (approved):
LGTM. Yes, task_class should be subtypes of Type[gokart.TaskOnKart]. :cool:

PR #291, opened 2022-07-02
Added should_dump_supplementary_log_files option

mski-iksm:
[What]
TaskOnKart.should_dump_supplementary_log_files is an option to control whether to dump supplementary files (task_log, random_seed, task_params, processing_time, module_versions) or not. Default is True, which means to dump as before.

[Why]
When each task runs, they will make 1 or more output files as defined in task.run(). Besides, 5 supplementary files (task_log, random_seed, task_params, processing_time, module_versions) will also be dumped. 
When using cloud data storage and running a large number of tasks, number of files will affect the price for data processing. This option will skip the dumping process of supplementary files to keep the price low.

Please review.

### DIFF
--- a/docs/task_on_kart.rst
+++ b/docs/task_on_kart.rst
@@ -244,3 +244,9 @@ TaskOnKart.fail_on_empty_dump
 Please refer to :doc:`for_pandas`.
 
 
+TaskOnKart.should_dump_supplementary_log_files
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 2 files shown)


hirosassa (changes requested):
Your changes looks good!

Please document this feature.
https://github.com/m3dev/gokart/blob/master/docs/task_on_kart.rst

mski-iksm:
@hirosassa Thanks for the comment. I've added a document!

hirosassa (approved):
LGTM!

kitagry (approved):
LGTM

hirosassa:
@mski-iksm Thanks again for your contribution!

PR #294, opened 2022-07-21
drop Python 3.7 support

hirosassa:
SSIA

Dropped Python 3.7 support because `pandas` and `numpy` dropped it and [this](https://endoflife.date/python).


mski-iksm (approved):
LGTM!

hirosassa:
Thanks for your reviews!

PR #295, opened 2022-07-30
Hide private params from logs

ujiuji1259:
I've implemented `__str__` for task representations with only public parameters.

## Problems
When running the task with [private parameters](https://luigi.readthedocs.io/en/stable/parameters.html#parameter-visibility), the worker now emits the logs including the raw values of private parameters.

This is caused by the implementation of `__repr__` that builds the task representation with all the significant parameters, and `str(task)` calls it because `__str__` is not implemented. It is reasonable that `__repr__` Includes all the siginificant parameters since `__repr__` should identify the task,  but I think `str(task)` should not include private parameters.

## Solution
I've implemented `_get_task_string(only_public=False)` and `__str__`, and added unittest for them.

### DIFF
--- a/gokart/task.py
+++ b/gokart/task.py
@@ -8,6 +8,7 @@ from typing import Any, Callable, Dict, List, Optional, Set, Union
 
 import luigi
 import pandas as pd
+from luigi.parameter import ParameterVisibility
 
 import gokart
 from gokart.file_processor import FileProcessor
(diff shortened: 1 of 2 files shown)


hirosassa (approved):
@ujiuji1259 Thank you for your contribution! LGTM!

mski-iksm (approved):
@ujiuji1259 
LGTM! Thank you for the contribution!

PR #299, opened 2022-10-31
set columns with list

mski-iksm:
Problem: gokart could not use pandas>=1.5.0

In pandas>=1.5.0, columns cannot be declared with set().
https://pandas.pydata.org/docs/whatsnew/v1.5.0.html#other-api-changes

However, gokart.TaskOnKart.required_columns will use set() to build pd.DataFrame when required_column is set with set() and pd.DataFrame is empty.

I changed the code to convert set to list and then build pd.DataFrame.

### DIFF
--- a/gokart/s3_zip_client.py
+++ b/gokart/s3_zip_client.py
@@ -17,6 +17,9 @@ class S3ZipClient(ZipClient):
 
     def make_archive(self) -> None:
         extension = os.path.splitext(self._file_path)[1]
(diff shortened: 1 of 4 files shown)


hirosassa:
Your changes looks good 👍  Please check the CI running on Python 3.10.

kitagry (approved):

hirosassa (approved):

Issue #268, opened 2022-01-19
[Feature Request] Runtime validation of `TaskInstanceParameter()` by subclass bound


ganow:
Hello, thank you for developing such a great tool!

## Summary

I have one feature request to validate `TaskInstanceParameter()` at runtime by its subclass bound like the following:

```python
class UpstreamBase(gokart.TaskOnKart): ...
class UpstreamA(UpstreamBase): ...
class UpstreamB(UpstreamBase): ...

class Example(gokart.TaskOnKart):
    upstream_task = gokart.TaskInstanceParameter(bound=UpstreamBase)
    ...
    def requires(self):
        return self.upstream_task  # guaranteed it is a subclass of UpstreamBase at runtime
    ...
```

## Detail

More concrete motivating example is as follows. Suppose we want to perform some feature embedding pre-processing followed by actual data analysis task on top of it. Since we would like to empirically compare which embedding method is better, consider making this pre-processing task (upstream task) abstract and injecting actual choice as a `TaskInstanceParameter()`. The examples of illustrative pipelines are like:
- pattern 1: data -> **PCA** embedding -> downstream analysis (classification, visualization, etc.)
- pattern 2: data -> **Isomap** embedding -> downstream analysis (classification, visualization, etc.)

In such a situation, we may want to limit the task instance parameter to be injected as an upstream task to some tasks with a specific output format, rather than all possible tasks defined. The current default behavior of `TaskInstanceParameter()` can cause potential bugs because it does not raise any kind of exceptions in all subclasses of `luigi.Task()`.

Here is an example code:

```python
from abc import ABCMeta, abstractmethod

import gokart
import luigi

import numpy as np
from sklearn.decomposition import PCA
from sklearn.manifold import Isomap

class CreateFakeData(gokart.TaskOnKart):
    def run(self):
        self.dump(np.random.normal(size=(100, 20)))

class EmbedTask(gokart.TaskOnKart, metaclass=ABCMeta):
    data_task = CreateFakeData()
    n_components = luigi.IntParameter(default=2)

    def requires(self):
        return self.data_task

    def run(self):
        data = self.load()
        embedding = self.get_embedding(data)
        assert embedding.shape == (len(data), self.n_components)
        self.dump(embedding)

    @abstractmethod
    def get_embedding(self, data): ...

class PCAEmbed(EmbedTask):
    whiten = luigi.BoolParameter(default=False)

    def get_embedding(self, data):
[...]

PR #305, opened 2023-02-04
Feature/ Runtime validation of TaskInstanceParameter() and ListTaskInstanceParameter() by subclass bound

ujiuji1259:
Issue: #268 

I've implemented runtime validation of TaskInstanceParameter() and ListTaskInstanceParameter() by subclass bound.

@ganow I'd like to use this feature, so I've created a PR based on your idea. Thank you for your feature request and the concrete implementation idea!

### DIFF
--- a/gokart/parameter.py
+++ b/gokart/parameter.py
@@ -12,11 +12,12 @@ logger = getLogger(__name__)
 
 class TaskInstanceParameter(luigi.Parameter):
 
-    def __init__(self, *args, **kwargs):
-        expected_type = kwargs.pop('expected_type', gokart.TaskOnKart)
-        if isinstance(expected_type, type):
+    def __init__(self, expected_type=None, *args, **kwargs):
+        if expected_type is None:
+            self.expected_type = gokart.TaskOnKart
+        elif isinstance(expected_type, type):
             self.expected_type = expected_type
-        else:
+        elif expected_type is not None:
             raise TypeError(f'expected_type must be a type, not {type(expected_type)}')
         super().__init__(*args, **kwargs)
 
@@ -64,11 +65,12 @@ class _TaskInstanceEncoder(json.JSONEncoder):
 
 class ListTaskInstanceParameter(luigi.Parameter):
 
-    def __init__(self, *args, **kwargs):
-        expected_type = kwargs.pop('expected_type', gokart.TaskOnKart)
-        if isinstance(expected_type, type):
+    def __init__(self, expected_type=None, *args, **kwargs):
+        if expected_type is None:
+            self.expected_type = gokart.TaskOnKart
+        elif isinstance(expected_type, type):
             self.expected_type = expected_type
-        else:
+        elif expected_type is not None:
             raise TypeError(f'expected_type must be a type, not {type(expected_type)}')
         super().__init__(*args, **kwargs)
 
--- a/gokart/parameter.py
+++ b/gokart/parameter.py
@@ -12,11 +12,12 @@ logger = getLogger(__name__)
 
 class TaskInstanceParameter(luigi.Parameter):
 
-    def __init__(self, bound=gokart.TaskOnKart, *args, **kwargs):
-        if isinstance(bound, type):
-            self._bound = bound
+    def __init__(self, *args, **kwargs):
+        expected_type = kwargs.pop('expected_type', gokart.TaskOnKart)
+        if isinstance(expected_type, type):
+            self.expected_type = expected_type
         else:
-            raise ValueError(f'bound must be a type, not {type(bound)}')
+            raise TypeError(f'expected_type must be a type, not {type(expected_type)}')
         super().__init__(*args, **kwargs)
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


ganow:
@mski-iksm Thank you for creating the PR!

I forgot to make a PR after making this issue.

ujiuji1259:
@mski-iksm @hirosassa @Hi-king please review!

hirosassa (approved):
Looks good!

gokart/parameter.py
```diff
@@ -5,11 +5,21 @@
...
 
+import gokart
+
 logger = getLogger(__name__)
 
 
 class TaskInstanceParameter(luigi.Parameter):
 
+    def __init__(self, *args, **kwargs):
+        bound = kwargs.pop('bound', gokart.TaskOnKart)
```
Hi-king, line 16: If you mean subtype bound, how about just use 'type' keyword? There might be both upper and lower for 'bound' keyword.

gokart/parameter.py
```diff
@@ -36,6 +46,11 @@ def serialize(self, x):
         values = dict(type=x.get_task_family(), params=params)
         return luigi.DictParameter().serialize(values)
 
+    def normalize(self, v):
+        if not isinstance(v, self._bound):
+            raise ValueError(f'{v} is not an instance of {self._bound}')
```
Hi-king, line 51: How about using TypeError :)

gokart/parameter.py
```diff
@@ -5,11 +5,21 @@
...
 from luigi import task_register
 
+import gokart
+
 logger = getLogger(__name__)
 
 
 class TaskInstanceParameter(luigi.Parameter):
 
+    def __init__(self, *args, **kwargs):
```
Hi-king, line 15: Does following works as the same manner?

```
def __init__(self, bound=gokart.TaskOnKart, *args, **kwargs):
```

gokart/parameter.py
ujiuji1259: TypeError looks better! I'll fix it.

gokart/parameter.py
ujiuji1259: Thank you for the suggestion. 

It doesn't work due to the following error. Maybe it's because  the default value is evaluated in `__init__.py` before initializing `TaskOnKart` (TaskInstanceParameter is imported in `task.py`).

```AttributeError: partially initialized module 'gokart' has no attribute 'TaskOnKart' (most likely due to a circular import)```

**Comment by ujiuji1259** [comment]
> Thank you for the reviews!
> 
> I found [`_warn_on_wrong_param_type`](https://github.com/spotify/luigi/blob/ddd94d03cd67933c619007cd47e4774543e5c2b3/luigi/parameter.py#L292), and think overriding `_warn_on_wrong_param_type` to raise unexpected param type is more suitable in this situation.
> 
> How about this implementation?
> 
> Sample
> 
> ```python
> class TaskInstanceParameter(luigi.Parameter):
>     ...
> 
>     def _warn_on_wrong_param_type(self, param_name, param_value):
>         if self.__class__ != TaskInstanceParameter:
>             return
>         if not isinstance(param_value, self.expected_type):
>             raise TypeError(f'{param_value} is not an instance of {self.expected_type}')
> ```

**Comment by ujiuji1259** [comment]
I've updated implementation to use [_warn_on_wrong_param_type](https://github.com/spotify/luigi/blob/ddd94d03cd67933c619007cd47e4774543e5c2b3/luigi/parameter.py#L292).

Could you review again? @mski-iksm @hirosassa @Hi-king

File: gokart/parameter.py [hunk 4]
```diff
@@ -36,6 +47,12 @@ def serialize(self, x):
         values = dict(type=x.get_task_family(), params=params)
         return luigi.DictParameter().serialize(values)
 
+    def _warn_on_wrong_param_type(self, param_name, param_value):
+        if self.__class__ != TaskInstanceParameter:
```
mski-iksm, line 51: I think this return is for Paramteres who has inherited TaskInstanceParameter.

I think expected_type check is also important for those parameters, so removing this maybe considerable.

gokart/parameter.py
```diff
@@ -48,12 +65,28 @@ def default(self, obj):
 
 class ListTaskInstanceParameter(luigi.Parameter):
 
+    def __init__(self, expected_type=None, *args, **kwargs):
```
mski-iksm, line 68: expected_type is for the type of elements in the list.

So, I think expected_elements_type may be a better name.

mski-iksm (changes requested):

gokart/parameter.py
```diff
@@ -5,11 +5,22 @@
...
 
 
 class TaskInstanceParameter(luigi.Parameter):
 
+    def __init__(self, expected_type=None, *args, **kwargs):
+        if expected_type is None:
+            self.expected_type = gokart.TaskOnKart
+        elif isinstance(expected_type, type):
+            self.expected_type = expected_type
+        elif expected_type is not None:
```
Hi-king, line 20:
```suggestion
        else:
```

If there're no need for special care, just putting else can work.

gokart/parameter.py
ujiuji1259: Thank you for the comment! I've removed this class check

gokart/parameter.py
ujiuji1259: I've changed `expected_type` to `expected_elements_type`.

Hi-king (approved):
LGTM

mski-iksm (approved):
@ujiuji1259 LGTM!

Hi-king:
@ganow Thx for suggesting nice feature for improving reliability :)

@ujiuji1259 Thank you for the reasonable implementation! :+1:

PR #308, opened 2023-02-11
fix typo in Python version

hirosassa:
SSIA


hirosassa (commented):
Since this is the small fix-typo-PR, I'll merge this by my self.

PR #317, opened 2023-06-06
Set encoding when saving csv

mski-iksm:
I've added CsvFileProcessor._encoding to allow specifying encodings when dumping in csv.

PR #327, opened 2023-10-11
[docs] fix csv encoding option

ujiuji1259:
https://github.com/m3dev/gokart/pull/317 added the `encoding` option into `CsvFileProcessor`, but the following code in the [docs](https://github.com/m3dev/gokart/blob/master/docs/task_on_kart.rst#dump-csv-with-encoding) doesn't work.

```python
def output(self):
    return self.make_target('file_name.csv', encoding='cp932')
```

In fact, we have to directly pass `CsvFileProcessor` with `encoding='cp932'` to `TaskOnKart.make_target`.

### DIFF
--- a/docs/task_on_kart.rst
+++ b/docs/task_on_kart.rst
@@ -270,7 +270,7 @@ If you want to dump csv file with other encodings, you can use `encoding` parame
 
     from gokart.file_processor import CsvFileProcessor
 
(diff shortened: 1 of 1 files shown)


Hi-king:
@ujiuji1259 THX!

Hi-king (approved):

PR #333, opened 2023-11-09
Add complete_check_at_run parameter

mski-iksm:
## What is this?
TaskOnKart.complete_check_at_run is a parameter to control whether to check completed cache file existence at TaskOnKArt.run().
If complete_check_at_run==True, when the worker attempt to run the task, it will check the cache file existence, and if exists, run() will be skipped.

## Why is this needed?
In the original code, complete check is only evaluated at task scheduling, which is at the very beginning.
However, cache status may change when it comes to run().

This often happens when jobs in multiple nodes runs the same task.
When node1 and node2 runs the same task, the first finished node will create the completed cache, but the other node may rerun the task and recreate the same cache file.

It is okay to run same task again, but it is not efficient.
To prevent running the completed task again, I've implemented TaskOnKart.complete_check_at_run option.

### DIFF
--- a/gokart/task.py
+++ b/gokart/task.py
@@ -77,8 +77,9 @@ class TaskOnKart(luigi.Task):
         description='Whether to dump supplementary files (task_log, random_seed, task_params, processing_time, module_versions) or not. \
          Note that when set to False, task_info functions (e.g. gokart.tree.task_info.make_task_info_as_tree_str()) cannot be used.',
         significant=False)
-    complete_check_at_run: bool = ExplicitBoolParameter(
-        default=False, description='Check if output file exists at run. If exists, run() will be skipped.', significant=False)
+    complete_check_at_run: bool = ExplicitBoolParameter(default=False,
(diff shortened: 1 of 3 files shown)


gokart/task.py
```diff
@@ -86,6 +90,9 @@ def __init__(self, *args, **kwargs):
         self._rerun_state = self.rerun
         self._lock_at_dump = True
 
+        if self.complete_check_at_run:
+            self.run = task_complete_check_wrapper(run_func=self.run, complete_check_func=self.complete)
```
yokomotod, line 94: @mski-iksm  What will the log look like?

I'm wondering if we should add a log that shows what I skipped.

mski-iksm: @yokomotod 
I've added a log as following: 
https://github.com/m3dev/gokart/pull/333/files#diff-1c658ac433692b62825bf8604431590512f2ae636030e2041aa457858b486aaeR11

Hi-king (approved):

Hi-king:
@mski-iksm THX!

PR #335, opened 2023-11-17
gokart.build(reset_register=False) can use PandasTypeConfig

Hi-king:
By default, gokart.build() resets Register. So PandasTypeConfig is also ignored.

This PR keeps PandasTypeConfig even if reset_register=True.


kitagry (approved):

mski-iksm:
LGTM!

PR #336, opened 2023-11-21
add __name__ to RunWithLock

mski-iksm:


gokart/run_with_lock.py
```diff
@@ -16,6 +16,9 @@ def __call__(self, instance):
     def __get__(self, instance, owner_class):
         return partial(self.__call__, instance)
 
+    def __name__(self):
+        return self._func.__name__
+
```
Hi-king, line 21: How about using dynamic wrapper pattern ?
https://python-patterns.guide/gang-of-four/decorator-pattern/#implementing-dynamic-wrapper

```suggestion
    def __getattr__(self, name):
        return getattr(self._func, name)

```

PR #337, opened 2023-11-21
added @functools.wraps(run_func) to task_complete_check_wrapper

mski-iksm:
I've added @functools.wraps not to change wrapped function's name.

### DIFF
--- a/gokart/task_complete_check.py
+++ b/gokart/task_complete_check.py
@@ -1,3 +1,4 @@
+import functools
 from logging import getLogger
 from typing import Callable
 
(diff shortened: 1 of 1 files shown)


Hi-king:
LGTM!

yokomotod:
@mski-iksm Can we have any kind of unit test ?

I'm not sure what problem happen if decorate function name changed. (but yes, sounds problem 😅

mski-iksm:
@yokomotod 
This PR is not to fix any bug, instead it is for readability.

So there is no concrete problem case, and therefore I think no test is required.

yokomotod:
I see, then I agree to not have unit test for `@functools.wraps` functionality itself !

and +1 for using `@functools.wraps` habit

yokomotod (approved):

Hi-king (approved):

PR #342, opened 2024-01-10
set TaskOnKart.raise_task_lock_exception_on_collision as false

mski-iksm:
RedisParams.raise_task_lock_exception_on_collision is a option whether to fail when TaskLockException is raised (when True), or just wait until the lock is released (when False).

Originally, this setting was passed from TaskOnKart.raise_task_lock_exception_on_collision.
However, this option is not needed since task collision lock of TaskOnKart.dump() don't want to fail even when TaskLockException is raised.
Rather, we want to wait until the lock is released.

In this PR, I've changed the raise_task_lock_exception_on_collision as False for task collision lock of TaskOnKart.dump().


gokart/task.py
```diff
@@ -176,7 +173,8 @@ def make_target(self, relative_file_path: Optional[str] = None, use_unique_id: b
                                          redis_host=self.redis_host,
                                          redis_port=self.redis_port,
                                          redis_timeout=self.redis_timeout,
-                                         raise_task_lock_exception_on_collision=self.raise_task_lock_exception_on_collision)
+                                         raise_task_lock_exception_on_collision=False)
```
Hi-king, line 176: @mski-iksm Plz describe why we need to keep  `raise_task_lock_exception_on_collision` for make_redis_params while this is always set to be False

mski-iksm: @Hi-king 
In the coming feature, TaskOnKart.run() will be wrapped with task collision lock to prevent running same task at the same time.
I want to keep RedisParams.raise_task_lock_exception_on_collision to be used in this new feature.

Hi-king: @mski-iksm Thx for clarifying :)

Hi-king (approved):

PR #357, opened 2024-03-06
relax python upper constraint

yokomotod:
`numpy` seems un-pinned python upper version https://github.com/numpy/numpy/pull/24962 so now we can relax constraint too.


pyproject.toml
```diff
@@ -16,7 +16,7 @@ style = "pep440"
 pattern = "^(?P<base>\\d+\\.\\d+\\.\\d+)"
 
 [tool.poetry.dependencies]
-python = ">=3.9,<3.13"
+python = ">=3.9,<4"
```
yokomotod, line 19: Some other dependencies have `<4` constraint, so `">=3.9"` causes problem (those libraries will downgrade to old version which has no `<4` constraint)

ex. `fakeredis` https://github.com/m3dev/gokart/blob/4e76952ddc1768c6a4978c3db225c434424a3ca2/poetry.lock#L504

Hi-king (approved):
Confirmed no critical diff on` poetry.lock` . THX!

PR #359, opened 2024-03-21
bugfix: FeatherFileProcessor set wrong index

kitagry:
When original DataFrame doesn't set index name, FeatherFileProcessor set index_name as 'None'. 

### Why we don't fix `dump` not to dump index None?

It is important for the compatibility, the already dumped feather file has the name `__feather_gokart_index__None`. And I think no one doesn't set `"None"` name index. :wink:

### DIFF
--- a/gokart/file_processor.py
+++ b/gokart/file_processor.py
@@ -247,6 +247,7 @@ class FeatherFileProcessor(FileProcessor):
             index_column_name = f'{self.INDEX_COLUMN_PREFIX}{dump_obj.index.name}'
             assert index_column_name not in dump_obj.columns, f'column name {index_column_name} already exists in dump_obj. \
                 Consider not saving index by setting store_index_in_feather=False.'
+            assert dump_obj.index.name != 'None', 'index name is "None", which is not allowed in gokart. Consider setting another index name.'
 
             dump_obj[index_column_name] = dump_obj.index
             dump_obj = dump_obj.reset_index(drop=True)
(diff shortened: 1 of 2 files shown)


gokart/file_processor.py
```diff
@@ -245,6 +247,7 @@ def dump(self, obj, file):
             index_column_name = f'{self.INDEX_COLUMN_PREFIX}{dump_obj.index.name}'
             assert index_column_name not in dump_obj.columns, f'column name {index_column_name} already exists in dump_obj. \
                 Consider not saving index by setting store_index_in_feather=False.'
+            assert dump_obj.index.name != 'None', 'index name is None. Consider setting index name.'
```
yokomotod, line 250: [imo]

to make sure we means str "None",

```suggestion
            assert dump_obj.index.name != 'None', 'index name is `"None"`, which is not allowed in gokart. Consider setting another index name.'
```

test/test_file_processor.py
```diff
@@ -65,3 +65,49 @@ def test_load_csv_with_cp932(self):
                 # read with cp932 to check if the file is dumped with cp932
                 loaded_df = processor.load(f)
                 pd.testing.assert_frame_equal(df, loaded_df)
+
+
+class TestFeatherFileProcessor(unittest.TestCase):
+    def test_feather_should_return_same_dataframe(self):
+        df = pd.DataFrame({'あ': [1, 2, 3], 'い': [4, 5, 6]})
```
yokomotod, line 72: [question][nits]

any reason you use hiragana here ?

it doesn't depends on it's hiragana/ascii right ? then ascii sounds fine (to avoid question that hiragana is important or not)

yokomotod (approved):
LGTM with nits comments

mski-iksm:
LGTM! Thank you for bugfix!

PR #360, opened 2024-03-22
Draft: Dump with Type Check 

dn070017:
dump data with type


hirosassa:
@dn070017 Please check mypy failure.

gokart/target.py
```diff
@@ -91,9 +93,14 @@ def _get_task_lock_params(self) -> TaskLockParams:
 
     def _load(self) -> Any:
         with self._target.open('r') as f:
-            return self._processor.load(f)
+            obj = self._processor.load(f)
+            self._validator(obj)
```
Hi-king, line 97: @ujiuji1259 You can use retrun value of validator and let this process fail if return value is False

PR #361, opened 2024-03-22
Enable feather and parquet in S3

mamo3gr:
SSIA


hirosassa (approved):
Looks good! except for todo.

gokart/file_processor.py
```diff
@@ -222,10 +223,10 @@ def __init__(self, store_index_in_feather: bool):
         self.INDEX_COLUMN_PREFIX = '__feather_gokart_index__'
 
     def format(self):
-        return None
+        return luigi.format.Nop
 
     def load(self, file):
-        loaded_df = pd.read_feather(file.name)
+        loaded_df = pd.read_feather(BytesIO(file.read()))
```
yokomotod, line 229: @mamo3gr 

Can we avoid `file.read()` here? I think it will read all content onto memory.

I hope `BytesIO` accept `file` (reader-like) itself

mamo3gr: >I hope BytesIO accept file (reader-like) itself

Unfortunately it seems not to do so.

https://docs.python.org/3/library/io.html#binary-i-o

>Binary I/O (also called buffered I/O) expects bytes-like objects and produces bytes objects.

mamo3gr: @yokomotod 

It is a little bit heavy to enable streaming read because we need to modify codes in luigi. I leave FIXME comment there in 7abbc331d5f5a0f04f1e688e2dcc00ee076a0bb6. Can we merge this PR with these comments?

mamo3gr:
> add tests for securing behavior with GCS

I tried it and found to need the time. I wrote down the details to https://github.com/m3dev/gokart/issues/363 and would proceed to merge. May I do so?

gokart/file_processor.py
mamo3gr: Additionally I added a conditional branching with respect to whether the passed `file` is reader-like or not in 65bb5b0cd73e8250c6969d2f8a485368ccead2b5.

Hi-king (approved):

gokart/file_processor.py
mamo3gr: reverted the way of loading except S3 in 0a6880f1b9e6239bf92748235cc5ee62e05543cc.

yokomotod (approved):
LGTM!

PR #362, opened 2024-04-01
add type aware flatten method instead of `luigi.task.flatten`

kitagry:
Currently, we use `luigi.task.flatten`, but this method is not type aware (it return type is always `Unknown`. Because of this, we wrote missed type [here](https://github.com/m3dev/gokart/blob/837742d/gokart/tree/task_info_formatter.py?plain=1#L15) ( it should be `List[str]`, but `List[TaskOnKart]` ). 

So, I introduce type aware flatten method instead of `luigi.task.flatten`

### DIFF
--- a/gokart/utils.py
+++ b/gokart/utils.py
@@ -38,6 +38,8 @@ def flatten(targets: FlattenableItems[T]) -> list[T]:
         ['foo']
         >>> flatten(42)
         [42]
+
+    This method is copied and modified from [luigi.task.flatten](https://github.com/spotify/luigi/blob/367edc2e3a099b8a0c2d15b1676269e33ad06117/luigi/task.py#L958) in accordance with [Apache License 2.0](https://github.com/spotify/luigi/blob/367edc2e3a099b8a0c2d15b1676269e33ad06117/LICENSE).
     """
     if targets is None:
         return []
--- a/gokart/utils.py
+++ b/gokart/utils.py
@@ -1,7 +1,8 @@
 from __future__ import annotations
 
 import os
-from typing import Iterable, TypeAlias, TypeVar
+import sys
+from typing import Iterable, TypeVar
 
 import luigi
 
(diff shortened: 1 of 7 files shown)


kitagry:
TypeAlias is not support in Python3.9 😮‍💨

gokart/utils.py
```diff
@@ -7,3 +11,48 @@ def add_config(file_path: str):
     _, ext = os.path.splitext(file_path)
     luigi.configuration.core.parser = ext
     assert luigi.configuration.add_config_path(file_path)
+
+
+T = TypeVar('T')
+if sys.version_info >= (3, 10):
+    from typing import TypeAlias
+
+    FlattableItems: TypeAlias = T | Iterable['FlattableItems[T]'] | dict[str, 'FlattableItems[T]']
```
kitagry, line 20: TypeAlias is required in mypy. from Python 3.12, this will be replaced [`type statement`](https://docs.python.org/ja/3/reference/simple_stmts.html#type)

gokart/utils.py
```diff
@@ -7,3 +11,48 @@ def add_config(file_path: str):
...
+else:
+    from typing import Union
+
+    FlattableItems = Union[T, Iterable['FlattableItems[T]'], dict[str, 'FlattableItems[T]']]
```
kitagry, line 24: TypeAlias is not supported in Python 3.9. And, `|` is not supported.

gokart/utils.py
```diff
@@ -7,3 +11,48 @@ def add_config(file_path: str):
...
+
+
+def flatten(targets: FlattableItems[T]) -> list[T]:
```
Hi-king, line 27: @kitagry plz note this is slightly changed copy of luigi.task.flatten https://github.com/spotify/luigi/blob/367edc2e3a099b8a0c2d15b1676269e33ad06117/luigi/task.py#L958

gokart/utils.py
Hi-king, line 20: [nits] Flattenable(Flatten-able) might be more comprehensive

gokart/utils.py
Hi-king: How about testing working the same way as luigi.task.flatten?



``
self.assertEqual(sorted(luigi.task.flatten({'a': 'foo', 'b': 'bar'})), sorted(gokart.utils.flatten({'a': 'foo', 'b': 'bar'})))
``

kitagry: add note [57511ea](https://github.com/m3dev/gokart/pull/362/commits/57511ea3b4e970e854a69802824a192dd3ca4295)

gokart/utils.py
kitagry: fixed [d58fe04](https://github.com/m3dev/gokart/pull/362/commits/d58fe0495043b9f2d26032e7327872fd29731c4b)

gokart/utils.py
kitagry: I added test. (I think it is better to use concrete values instead of comparing to `luigi.task.flatten`)

mski-iksm (approved):
@kitagry LGTM!

hirosassa (approved):
Looks good!

PR #366, opened 2024-04-08
bugfix: output_paths type is wrong

kitagry:
In the following code, output_path is converted from [`TargetOnKart.path()`](https://github.com/m3dev/gokart/blob/master/gokart/target.py?plain=1#L44) which returns str. I'll trying to avoid this error with strict type in https://github.com/m3dev/gokart/pull/362, but this is a bit difficult for Python 3.9.

https://github.com/m3dev/gokart/blob/master/gokart/tree/task_info_formatter.py?plain=1#L68


tkda-h3:
LGTM

tkda-h3 (approved):

Hi-king:
LGTM!

PR #368, opened 2024-04-15
add TaskOnKart dump type

kitagry:
Add type to TaskOnKart. we can set type.

```python
class MyTask(gokart.TaskOnKart[str]):
    def run(self):
        self.dump(1)  # mypy error: expected str but got int
        self.dump('1')  # mypy ok
```

### DIFF
--- a/test/test_build.py
+++ b/test/test_build.py
@@ -1,7 +1,14 @@
 import logging
 import os
+import sys
 import unittest
 from copy import copy
+from typing import Dict
+
+if sys.version_info >= (3, 11):
+    from typing import assert_type
+else:
+    from typing_extensions import assert_type
 
 import luigi
 import luigi.mock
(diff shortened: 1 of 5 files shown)


test/test_build.py
```diff
@@ -79,6 +86,7 @@ def test_read_config(self):
         config_file_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'config', 'test_config.ini')
         gokart.utils.add_config(config_file_path)
         output = gokart.build(_DummyTask(), reset_register=False)
+        assert_type(output, str)
```
hirosassa, line 89: 👍 👍 👍 👍

hirosassa (approved):
LGTM!

hiro-o918 (approved):
LGTM

PR #376, opened 2024-05-21
setting priority random

mski-iksm:
Changing priority randomly to sort the order of task scheduling randomly.

This is useful when jobs are running in parallel.
Together with the complete check function during run (complete_check_at_run; default True), it will speeds up parallel execution, because each job will run different task in different order and same tasks will not be ran.

The disadvantage is that the order of execution is scattered, so it changes a little each time when looking at the log.

### DIFF
--- a/gokart/task.py
+++ b/gokart/task.py
@@ -1,6 +1,7 @@
 import hashlib
 import inspect
 import os
+import random
 import types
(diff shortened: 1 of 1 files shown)


Hi-king (approved):
Looks Good. Drawback should be small enough

PR #381, opened 2024-07-11
Load DataFrame cache with backward compatibility

ujiuji1259:
I have updated `PickleFileProcessor` to be able to load dataframes dumped in backward pandas versions.

## background
Because the pandas dataframe dumped by pickle is not backward compatible, TargetOnKart cannot load the cache of dataframes dumped in backward pandas version.
For example, TargetOnKart in pandas 2.1.0 cannot load the large dataframe dumped in pandas 1.5.3 (error: `ModuleNotFoundError: No module named 'pandas.core.indexes.numeric'`).

## Solution
In fact,  `pd.read_pickle` can load the dataframe dumped by lower pandas version by patching [`pickle.load`](https://github.com/pandas-dev/pandas/blob/d9cdd2ee5a58015ef6f4d15c7226110c9aab8140/pandas/compat/pickle_compat.py#L212) in pandas, so we can avoid the problem by using `pd.read_pickle` instead of `dill.load`.
`pd.read_pickle` can also load any objects other than dataframes, even the objects dumped by dill, we can naturally use it in `PickleFileProcessor`.

### DIFF
--- a/gokart/file_processor.py
+++ b/gokart/file_processor.py
@@ -14,6 +14,7 @@ import pandas.errors
 from luigi.format import TextFormat
 
 from gokart.object_storage import ObjectStorage
+from gokart.utils import load_dill_with_pandas_backward_compatibility
 
 logger = getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


Hi-king:
@ujiuji1259 Nice API survey & documentation :) LGTM

Hi-king (approved):

PR #382, opened 2024-07-12
fix load_dill_with_pandas_backward_compatibility() bug

mski-iksm:
When `pd.read_pickle(file)` failed, load_dill_with_pandas_backward_compatibility() will fall back and use `dill.load(file)`.

But before loading with dill, file cursor must be rewind to the beginning.

### DIFF
--- a/gokart/file_processor.py
+++ b/gokart/file_processor.py
@@ -82,7 +82,7 @@ class PickleFileProcessor(FileProcessor):
         return luigi.format.Nop
 
     def load(self, file):
-        if not ObjectStorage.is_buffered_reader(file):
+        if not ObjectStorage.is_buffered_reader(file) or not file.seekable():
             # we cannot use dill.load(file) because ReadableS3File does not have 'readline' method
             return load_dill_with_pandas_backward_compatibility(BytesIO(file.read()))
         return load_dill_with_pandas_backward_compatibility(_ChunkedLargeFileReader(file))
--- a/gokart/file_processor.py
+++ b/gokart/file_processor.py
@@ -82,8 +82,8 @@ class PickleFileProcessor(FileProcessor):
         return luigi.format.Nop
 
     def load(self, file):
-        if not ObjectStorage.is_buffered_reader(file) or not file.seekable():
-            # we cannot use dill.load(file) because ReadableS3File does not have 'readline' method
+        if not file.seekable():
+            # ReadableS3File is not seekable, therefor we need to wrap with BytesIO
             return load_dill_with_pandas_backward_compatibility(BytesIO(file.read()))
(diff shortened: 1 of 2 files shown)


Hi-king:
@mski-iksm Thx for pointing out a critical bug. Could you tell me which commit do you think this bug was introduced in?

Hi-king:
I hope a fixed test will be added to `test/test_file_processor.py` 🙏

gokart/utils.py
```diff
@@ -76,4 +78,5 @@ def load_dill_with_pandas_backward_compatibility(file: FileLike) -> Any:
     try:
         return pd.read_pickle(file)
     except Exception:
+        file.seek(0)
```
Hi-king, line 81: https://github.com/search?q=repo%3Aspotify%2Fluigi+seekable&type=code 
Some processors are not seekable. 
Plz check `file.seekable()` here and commenting limitations.

kitagry:
LGTM other than comments

mski-iksm:
> @mski-iksm Thx for pointing out a critical bug. Could you tell me which commit do you think this bug was introduced in?

This bug was introduced in the following PR

https://github.com/m3dev/gokart/pull/381

ujiuji1259 (approved):

hirosassa (approved):

gokart/file_processor.py
```diff
@@ -82,7 +82,7 @@ def format(self):
         return luigi.format.Nop
 
     def load(self, file):
-        if not ObjectStorage.is_buffered_reader(file):
+        if not ObjectStorage.is_buffered_reader(file) or not file.seekable():
             # we cannot use dill.load(file) because ReadableS3File does not have 'readline' method
             return load_dill_with_pandas_backward_compatibility(BytesIO(file.read()))
         return load_dill_with_pandas_backward_compatibility(_ChunkedLargeFileReader(file))
```
kitagry, line 88: `_ChunkedLargeFileReader` is not seekable. So, it shouldn't use load_dill_with_pandas_backward_compatibility.

gokart/utils.py
mski-iksm: @Hi-king 
I've added seeable assertion. Due to https://github.com/m3dev/gokart/pull/382/files#diff-d30054885311f8eb77523063db114b013d94dd98d39a2d4e34a6f3ec5f461bb1R85 , non-seekable file should not come here but I've added assertion in case.

Hi-king: @mski-iksm Perfect! THX!

Hi-king (approved):

Hi-king:
@mski-iksm THX for quick and careful bugfix :)

PR #391, opened 2024-08-30
Add TaskOnKart load type

kitagry:
Add load type for TaskOnKart.

```python
class SampleTask(gokart.TaskOnKart):
    str_task: gokart.TaskOnKart[str] = gokart.TaskInstanceParameter()
    int_task: gokart.TaskOnKart[int] = gokart.TaskInstanceParameter()

    def requires(self):
        return dict(str=self.str_task, int=self.int_task)

    def run(self):
        s = self.load(self.str_task)  # This type is inferred with "str"
        i = self.load(self.int_task)  # This type is inferred with "int"

SampleTask(
    str_task=StrTask(),  # mypy ok
    int_task=StrTask(),  # mypy error: Argument "int_task" to "StrTask" has incompatible type "StrTask"; expected "TaskOnKart[int]
)
```

TODO

- [x] should check TaskOnKart which is arguments of load method is in `requires` results
- [x] write docs

### DIFF
--- a/gokart/task.py
+++ b/gokart/task.py
@@ -385,6 +385,8 @@ class TaskOnKart(luigi.Task, Generic[T]):
             result: FlattenableItems[TargetOnKart] = input[target]
             return result
         if isinstance(target, TaskOnKart):
+            requires_unique_ids = [task.make_unique_id() for task in flatten(self.requires())]
+            assert target.make_unique_id() in requires_unique_ids, f'{target} should be in requires method'
             return target.output()
         return target
 
--- a/gokart/task.py
+++ b/gokart/task.py
@@ -26,6 +26,7 @@ logger = getLogger(__name__)
 
 
 T = TypeVar('T')
+K = TypeVar('K')
 
 
 class TaskOnKart(luigi.Task, Generic[T]):
@@ -272,7 +273,13 @@ class TaskOnKart(luigi.Task, Generic[T]):
             task_lock_params=task_lock_params,
         )
 
-    def load(self, target: Union[None, str, TargetOnKart] = None) -> Any:
+    @overload
+    def load(self, target: Union[None, str, TargetOnKart] = None) -> Any: ...
+
+    @overload
+    def load(self, target: 'TaskOnKart[K]') -> K: ...
+
+    def load(self, target: Union[None, str, TargetOnKart, 'TaskOnKart[K]'] = None) -> Any:
         def _load(targets):
             if isinstance(targets, list) or isinstance(targets, tuple):
                 return [_load(t) for t in targets]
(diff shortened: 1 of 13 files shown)


test/test_task_on_kart.py
```diff
@@ -2,8 +2,8 @@
 import pathlib
 import unittest
 from datetime import datetime
-from typing import Any, Dict
-from unittest.mock import MagicMock, patch
+from typing import Any, Dict, List, cast
+from unittest.mock import Mock, patch
```
kitagry, line 6: `isinstance(MagicMock, Iterable)` returns true. It is difficult to test, so I use `Mock` instead of `MagicMock`.

gokart/task.py
```diff
@@ -369,14 +385,20 @@ def _to_str_params(task):
...
         if isinstance(target, str):
             input = self.input()
             assert isinstance(input, dict), f'input must be dict[str, TargetOnKart], but {type(input)} is passed.'
             result: FlattenableItems[TargetOnKart] = input[target]
             return result
+        if isinstance(target, Iterable):
+            return [self._get_input_targets(t) for t in target]
+        if isinstance(target, TaskOnKart):
+            requires_unique_ids = [task.make_unique_id() for task in flatten(self.requires())]
+            assert target.make_unique_id() in requires_unique_ids, f'{target} should be in requires method'
```
hiro-o918, line 400: :+1:

hiro-o918 (commented):
[imo]
can you add some simple tests to `test_build.py` to check `self.load(self.task)` works?

It will be a good example to show this feature as well as check this these method works correct without mock.

mski-iksm:
@kitagry 
Thank you for suggesting great feature!

How about adding new document about this feature as in https://github.com/m3dev/gokart/pull/391#issue-2497559110 ?

mski-iksm (approved):
LGTM!

test/test_task_on_kart.py
```diff
@@ -191,23 +191,25 @@ def test_default_large_dataframe_target(self):
         task = _DummyTaskD()
         default_large_dataframe_target = task.make_large_data_frame_target()
         self.assertIsInstance(default_large_dataframe_target, ModelTarget)
-        self.assertEqual(f'_DummyTaskD_{task.task_unique_id}.zip', pathlib.Path(default_large_dataframe_target._zip_client._file_path).name)
+        target = cast(ModelTarget, default_large_dataframe_target)
```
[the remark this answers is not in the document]
kitagry: `assertIsInstance` doesn't cast type. So, we need to cast type separately.

hirosassa (approved):
LGTM!!

PR #392, opened 2024-08-31
fix is_parameter_call

kitagry:
Currently, luigi doesn't provide py.typed. it will be released next to 3.5.1.

https://github.com/spotify/luigi/pull/3297

With the following code, we can't assume type correctly.

```python                                                                      
from luigi import Parameter

    param = Parameter()
```

### DIFF
--- a/gokart/mypy.py
+++ b/gokart/mypy.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 import re
 from typing import Callable, Final, Iterator, Literal, Optional
 
+import luigi
 from mypy.expandtype import expand_type, expand_type_by_instance
 from mypy.nodes import (
     ARG_NAMED_OPT,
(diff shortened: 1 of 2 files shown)


hiro-o918 (approved):
LGTM!

gokart/mypy.py
```diff
@@ -413,6 +415,16 @@ def is_parameter_call(expr: Expression) -> bool:
...
         return PARAMETER_FULLNAME_MATCHER.match(type_info.fullname) is not None
+
+    # Currently, luigi doesn't provide py.typed. it will be released next to 3.5.1.
+    # https://github.com/spotify/luigi/pull/3297
+    # With the following code, we can't assume correctly.
+    #
+    # from luigi import Parameter
+    # class MyTask(gokart.TaskOnKart):
+    #     param = Parameter()
+    if isinstance(type_info, Var) and luigi.__version__ <= '3.5.1':
```
mski-iksm, line 426: maybe?

```suggestion
    if isinstance(type_info, Var) and luigi.__version__ < '3.5.1':
```

mski-iksm (approved):
LGTM!

gokart/mypy.py
kitagry: typed luigi is not yet released. So, I think it is correct.

PR #396, opened 2024-09-19
fix: handle default type for `DateMinuteParameter`

### LINKED ISSUE #395: mypy error: luigi.DateMinuteParameter
```python
datetime: datetime = luigi.DateMinuteParameter(interval=10, default=datetime.now())
```

got

```
error: Incompatible types in assignment (expression has type "int", variable has type "datetime")  [assignment]
```

hiro-o918:
## What
* fixes #395
* before this PR, `default` argument is assumed to be positioned on the first argument
  * but some paramters like `DateMinuteParmeter` not
* This PR aims to find the position of `default` argument by the keyword name

### DIFF
--- a/gokart/mypy.py
+++ b/gokart/mypy.py
@@ -97,8 +97,13 @@ class TaskOnKartPlugin(Plugin):
           foo: int = luigi.IntParameter()
           ```
         """
+        try:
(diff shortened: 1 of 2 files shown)


kitagry (approved):
LGTM!

Issue #293, opened 2022-07-21
Dict with length 1 loses its key after self.load()


naokimaejima:
Thanks for this useful module. Let me report a strange behavior.

One of my tasks dumps a dict with length 1 (like `InputTask` as an example below).
When I load the dumped intermediate `.pkl` file from the task , I can read it intact,
But when I `self.load()` it in another task, the key seems to be lost.

[pipeline example]

```python
class InputTask(gokart.TaskOnKart):

    def run(self):
        a_dict = {"a":1}
        # a_dict = {"a":1, "b":2}
        self.dump(a_dict)

    def output(self):
        return self.make_target("from_input_task.pkl")

class OutputTask(gokart.TaskOnKart):

    input_task = gokart.TaskInstanceParameter()

    def requires(self):
        return self.input_task

    def run(self):
        a_dict = self.load()
        self.dump(a_dict)

    def output(self):
        return self.make_target("from_output_task.pkl")

class MainTask(gokart.TaskOnKart):

    def requires(self):
        return OutputTask(input_task=InputTask())
```

When I load pickle file, the key of the dict is lost as below.

```python
>> python -m pickle resources/from_input_task_40b1decdc02b4e434a354c04860ed95a.pkl 
{'a': 1}
>> python -m pickle resources/from_output_task_5ece2af2be7b8ef168b704b0dd233858.pkl
1
```

As far as I know, this seems to happen only in the case of dict with length 1.
I guess that the codes below is the cause of the bug. I appreciate if you check it.

https://github.com/m3dev/gokart/blob/c5376e64d1787130c072e3cbda2b8b6e2620e8e3/gokart/task.py#L235-L236

nersonu:
@hirosassa @Hi-king
When parsing JSON, it is often the case that the dict root key has a length of 1.
It seems to me that if you want to include a process to distinguish between a dict that is the result of a task and a dict specified in requires, you would have to change luigi.
Is it conceivable to eliminate the L235-236 and L265-266 process?

mski-iksm:
@naokimaejima @nersonu 

Thank you for raising interesting issue.

> Is it conceivable to eliminate the L235-236 and L265-266 process?

I agree with your idea.



As @naokimaejima mentioned, this behavior is caused because of following code.

```
if target is None and isinstance(data, dict) and len(data) == 1: 
     return list(data.values())[0] 
```

This is implemented to make loading easier when using default requires().

``` 
class A(gokart.TaskOnKart):
[...]

PR #399, opened 2024-09-28
Breaking change: Don't extract value from 1-length dict

mski-iksm:
FIX: https://github.com/m3dev/gokart/issues/293

As mentioned in above issue, when loading 1-length dictionary with `self.load()` (no target specified), the value of dict will be loaded instead of dictionary itself.

Although this feature allows shorthand as described in the issue, the benefit is very limited and disadvantage caused by confusion is more significant.

Therefore, I want to delete this implementation.

----

This PR will cause following change.

```
class A(gokart.TaskOnKart):
    task_a = gokart.TaskInstanceParameter()

    def run(self):
        a = self.load()
        # From this PR, `a` will become a dictionary, instead of value of output of task_a
        # so, if you want to extract the value, you must specify key name as
        # a = self.load('task_a')
```

### DIFF
--- a/gokart/task.py
+++ b/gokart/task.py
@@ -291,8 +291,6 @@ class TaskOnKart(luigi.Task, Generic[T]):
             return targets.load()
 
         data = _load(self._get_input_targets(target))
-        if target is None and isinstance(data, dict) and len(data) == 1:
-            return list(data.values())[0]
         return data
 
     @overload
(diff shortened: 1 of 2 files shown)


gokart/task.py
```diff
@@ -291,8 +291,6 @@ def _load(targets):
             return targets.load()
 
         data = _load(self._get_input_targets(target))
-        if target is None and isinstance(data, dict) and len(data) == 1:
-            return list(data.values())[0]
         return data
```
hirosassa, line 294: how about followings:
```python
return _load(self._get_input_targets(target))
```

hirosassa (commented):
just nits comment

hirosassa (approved):
LGTM

kitagry (approved):

PR #401, opened 2024-10-13
feat: add tests to check tasks can be pickled

hiro-o918:


test/test_task_on_kart.py
```diff
@@ -657,5 +658,18 @@ def test_run_when_complete_check_at_run_is_true_and_task_is_completed(self):
         task.dump.assert_not_called()
 
 
+class TestPickleTaskOnKart:
+    def test_pickle_and_unpickle(self):
+        task = _DummyTask(redis_host='0.0.0.0', redis_port=12345, redis_timeout=180, should_lock_run=True)
+        pickled = pickle.dumps(task)
```
mski-iksm, line 664: @hiro-o918 
Since gokart uses `dill.dumps()`, I thought this may not happen in runtime.

If this test is changed to dill.dumps(), would it solve the problem that motivated this pull request?

gokart/conflict_prevention_lock/task_lock_wrappers.py
```diff
@@ -83,19 +83,18 @@ def wrapper(*args, **kwargs):
...
         scheduler = set_lock_scheduler(task_lock=task_lock, task_lock_params=task_lock_params)
 
         try:
             logger.debug(f'Task RUN lock of {task_lock_params.redis_key} locked.')
-            result = run_func()
+            run_func()
             task_lock.release()
             logger.debug(f'Task RUN lock of {task_lock_params.redis_key} released.')
             scheduler.shutdown()
-            return result
```
mski-iksm, line 98: Though I cannot remember, there must be some reason that this method has return, so I think we should reconsider before removing it.

test/test_task_on_kart.py
mski-iksm: I've understood. This was to reproduce the effect of following code.

```
gokart.build(_ParallelRunner(), reset_register=False, workers=20)
```

PR #402, opened 2024-10-14
feat: use gokart worker

hiro-o918:
## Background
* `gokart.TaskOnKart` wraps `run` function in order to inject some method when task run
  * But this makes unstable on pickling  like #401 
* This PR aims to add a custom worker for gokart instead of luigi

## What
* This PR just introduces a gokart worker which has same logic as luigi one

> [!NOTE]
> These changes does not affect any behaviors on default, because injecting the worker is an optional argument

## Detailed Changes
* [410aaf1](https://github.com/m3dev/gokart/pull/402/commits/410aaf1dabb32f82c2632c827f67d9e7d29cf52d)
  * copied worker from luigi and add test for worker
    * do not have to read worker.py of this commit
* [3e864a0](https://github.com/m3dev/gokart/pull/402/commits/3e864a030ba741f891e5398912b7d97deb0b03bc)
  * remove unused components and add type annotations
* [43e360d](https://github.com/m3dev/gokart/pull/402/commits/43e360d73317a7a5dcdded058206ba4b494ae894)
  * add a factory to inject our worker: https://github.com/spotify/luigi/blob/master/doc/running_luigi.rst#running-from-python-code

## Discussion

* I decided to add a new Worker to avoid inheritance
  * when inheriting it, we should overwrite some private functions. This makes our worker unstable.
* Please feel free to any comments for this way.

### DIFF
--- a/gokart/worker.py
+++ b/gokart/worker.py
@@ -30,37 +30,36 @@ ways between versions. The exception is the exception types and the
 
 import collections
 import collections.abc
+import contextlib
 import datetime
+import functools
 import getpass
 import importlib
+import json
 import logging
 import multiprocessing
 import os
-import signal
-import subprocess
-import sys
-import contextlib
-import functools
-
 import queue as Queue
 import random
+import signal
 import socket
+import subprocess
+import sys
 import threading
 import time
 import traceback
+from typing import Any, Dict, Generator, List, Literal, Optional, Set, Tuple
 
+import luigi
+import luigi.scheduler
+import luigi.worker
 from luigi import notifications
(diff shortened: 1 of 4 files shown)


gokart/worker.py
```diff
@@ -0,0 +1,1288 @@
+# -*- coding: utf-8 -*-
```
hiro-o918, line 1: You do not have to read this, because this file just copied

kitagry (approved):

hirosassa (approved):
LGTM

PR #403, opened 2024-10-25
feat: infer type of parameter when user doesn't specify type

kitagry:
Currently, we should write type hint for every parameters. This PR aims to infer type from Parameter names when user doesn't specify type.

## Before

```python
    foo = luigi.Parameter()

MyTask(foo=1)  # ok
```

## After

```python
    foo = luigi.Parameter()

MyTask(foo=1)  # Argument "foo" to "MyTask" has incompatible type "str"; expected "int"
```

### DIFF
--- a/gokart/mypy.py
+++ b/gokart/mypy.py
@@ -10,19 +10,16 @@ import re
 from typing import Callable, Final, Iterator, Literal, Optional
 
 import luigi
-from mypy.expandtype import expand_type, expand_type_by_instance
+from mypy.expandtype import expand_type
 from mypy.nodes import (
     ARG_NAMED_OPT,
-    ARG_POS,
     Argument,
     AssignmentStmt,
     Block,
     CallExpr,
     ClassDef,
-    Context,
     EllipsisExpr,
     Expression,
-    FuncDef,
     IfStmt,
     JsonDict,
     MemberExpr,
@@ -30,7 +27,6 @@ from mypy.nodes import (
     PlaceholderNode,
     RefExpr,
     Statement,
-    SymbolTableNode,
     TempNode,
     TypeInfo,
     Var,
@@ -45,13 +41,11 @@ from mypy.state import state
 from mypy.typeops import map_type_from_supertype
 from mypy.types import (
     AnyType,
-    CallableType,
     Instance,
     NoneType,
     Type,
     TypeOfAny,
     UnionType,
-    get_proper_type,
 )
 from mypy.typevars import fill_typevars
 
@@ -328,7 +322,7 @@ class TaskOnKartTransformer:
 
             current_attr_names.add(lhs.name)
             with state.strict_optional_set(self._api.options.strict_optional):
-                init_type = self._infer_task_on_kart_attr_init_type(sym, stmt)
+                init_type = sym.type
 
             # infer Parameter type
             if init_type is None:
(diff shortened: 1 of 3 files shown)


gokart/mypy.py
```diff
@@ -402,24 +407,115 @@ def _infer_task_on_kart_attr_init_type(self, sym: SymbolTableNode, context: Cont
...
+            base_type = self._get_type_from_args(parameter, 'enum')
+            if base_type is not None:
+                result_type = self._api.named_type('builtins.tuple', [base_type])
+
+        if result_type is None:
+            return None
+
+        # When parameter has Optional, it can be none value.
+        if 'Optional' in parameter_name:
+            result_type = UnionType([result_type, NoneType()])
```
hiro-o918, line 468: [nits]
```suggestion
            return UnionType([result_type, NoneType()])
```

gokart/mypy.py
```diff
@@ -402,24 +407,115 @@ def _infer_task_on_kart_attr_init_type(self, sym: SymbolTableNode, context: Cont
...
+    def _generate_type_from_parameters(self, parameter: Expression) -> Optional[Type]:
+        """
+        Generate default type from Parameter.
+        For example, when parameter is `luigi.parameter.Parameter`, this method should return `str` type.
+        """
+        parameter_name = _extract_parameter_name(parameter)
+        if parameter_name is None:
+            return None
+
+        result_type: Optional[Type] = None
```
hiro-o918, line 419: result_type is confused a little because it does not consider optional first 
so, why don't you rename `result_type` with `underlying_type`?

```suggestion
        underlying_type: Optional[Type] = None
```

Then, plz see the following review

gokart/mypy.py
```diff
@@ -329,6 +330,10 @@ def collect_attributes(self) -> Optional[list[TaskOnKartAttribute]]:
             with state.strict_optional_set(self._api.options.strict_optional):
                 init_type = self._infer_task_on_kart_attr_init_type(sym, stmt)
 
+            # infer Parameter type
+            if init_type is None:
+                init_type = self._generate_type_from_parameters(stmt.rvalue)
```
hiro-o918, line 335: please use consistent naming
```suggestion
                init_type = self._infer_type_from_parameters(stmt.rvalue)
```

gokart/mypy.py
```diff
@@ -329,6 +330,10 @@ def collect_attributes(self) -> Optional[list[TaskOnKartAttribute]]:
             with state.strict_optional_set(self._api.options.strict_optional):
                 init_type = self._infer_task_on_kart_attr_init_type(sym, stmt)
```
hiro-o918, line 331: [must]
currently `init_type` is inferred from `default` value.

But `_generate_type_from_parameters` should be prioritize to `default` type.

In addition, should we compare the type of default and inferred type from parameter class, though I do not know the way.

gokart/mypy.py
```diff
@@ -361,65 +360,115 @@ def _collect_parameter_args(self, expr: Expression) -> tuple[bool, dict[str, Exp
             return True, args
         return False, {}
 
-    def _infer_task_on_kart_attr_init_type(self, sym: SymbolTableNode, context: Context) -> Type | None:
```
hiro-o918, line 364: (just a memo for this change)
This function aims to infer type for pydantic field, which is not needed on gokart.

hiro-o918 (approved):
LGTM
Great changes!

gokart/mypy.py
```diff
@@ -361,65 +360,115 @@ def _collect_parameter_args(self, expr: Expression) -> tuple[bool, dict[str, Exp
...
+            base_type = self._get_type_from_args(parameter, 'var_type')
+            if base_type is not None:
+                underlying_type = self._api.named_type('builtins.tuple', [base_type])
+        elif parameter_name in ['luigi.parameter.EnumParameter']:
+            underlying_type = self._get_type_from_args(parameter, 'enum')
+        elif parameter_name in ['luigi.parameter.EnumListParameter']:
+            base_type = self._get_type_from_args(parameter, 'enum')
+            if base_type is not None:
+                underlying_type = self._api.named_type('builtins.tuple', [base_type])
+
```
hirosassa, line 415: memo: After Python 3.9 EOL, we can use `match case` syntax.

hirosassa (approved):
Looks great! Thanks!

PR #405, opened 2024-10-27
Check default value type of Parameter

kitagry:
Check default value of parameter.

```
import luigi
import gokart
    foo = luigi.IntParameter(default='s')  # Incompatible default for argument "foo" (default has type "str", argument has type "int")
```

### DIFF
--- a/test/test_mypy.py
+++ b/test/test_mypy.py
@@ -79,7 +79,6 @@ class MyEnum(enum.Enum):
     FOO = enum.auto()
 
 class MyTask(gokart.TaskOnKart):
-    # NOTE: mypy shows attr-defined error for the following lines, so we need to ignore it.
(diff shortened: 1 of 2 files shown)


test/test_mypy.py
```diff
@@ -79,7 +79,6 @@ class MyEnum(enum.Enum):
     FOO = enum.auto()
 
 class MyTask(gokart.TaskOnKart):
-    # NOTE: mypy shows attr-defined error for the following lines, so we need to ignore it.
```
hiro-o918, line 82: 👍

hiro-o918 (approved):

hiro-o918:
LGTM!!!!!

hiro-o918:
Oh, CI failed 😞

PR #406, opened 2024-11-08
feat: make worker skip running task when it is completed

hiro-o918:
## What
* This PR implements an alternative way to skip a task when it is completed just before run
  * This feature is currently provided by `complete_check_at_run` (#333)

## Commits
* 640310c: feat: add gokart_worker configurations as same as luigi one
* 20e852f:  feat: make worker skip run when a task is completed

## Why
* `complete_check_at_run` requires to overwrite `run` function of `TaskOnKart`, which causes the issue described by #401

## Note

This PR does not affect current features because `gokart.Worker` is not used anywhere.

### DIFF
--- a/gokart/worker.py
+++ b/gokart/worker.py
@@ -61,7 +61,9 @@ from luigi.task import DynamicRequirements, Task, flatten
 from luigi.task_register import TaskClassException, load_task
 from luigi.task_status import RUNNING
 
-logger = logging.getLogger('luigi-interface')
+from gokart.parameter import ExplicitBoolParameter
+
+logger = logging.getLogger(__name__)
 
 # Prevent fork() from being called during a C-level getaddrinfo() which uses a process-global mutex,
 # that may not be unlocked in child process, resulting in the process being locked indefinitely.
@@ -124,6 +126,7 @@ class TaskProcess(multiprocessing.Process):
         check_unfulfilled_deps: bool = True,
         check_complete_on_run: bool = False,
         task_completion_cache: Optional[Dict[str, Any]] = None,
+        skip_if_completed_pre_run: bool = True,
     ) -> None:
         super(TaskProcess, self).__init__()
         self.task = task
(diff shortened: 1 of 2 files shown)


gokart/worker.py
```diff
@@ -136,12 +139,19 @@ def __init__(
...
+        self.skip_if_completed_pre_run = skip_if_completed_pre_run
 
         # completeness check using the cache
         self.check_complete = functools.partial(luigi.worker.check_complete_cached, completion_cache=task_completion_cache)
 
+    def _run_task(self) -> Optional[collections.abc.Generator]:
+        if self.skip_if_completed_pre_run and self.check_complete(self.task):
+            logger.warning(f'{self.task} is skipped because the task is already completed.')
+            return None
+        return self.task.run()
```
hiro-o918, line 151: This line run a task

gokart/worker.py
```diff
@@ -136,12 +139,19 @@ def __init__(
         self.check_unfulfilled_deps = check_unfulfilled_deps
         self.check_complete_on_run = check_complete_on_run
         self.task_completion_cache = task_completion_cache
+        self.skip_if_completed_pre_run = skip_if_completed_pre_run
 
         # completeness check using the cache
         self.check_complete = functools.partial(luigi.worker.check_complete_cached, completion_cache=task_completion_cache)
 
+    def _run_task(self) -> Optional[collections.abc.Generator]:
+        if self.skip_if_completed_pre_run and self.check_complete(self.task):
```
hiro-o918, line 148: Here, a task is checked if completed to skip

gokart/worker.py
```diff
@@ -308,6 +318,68 @@ def run(self) -> None:
...
+        'each individual Task run.',
+    )
+    cache_task_completion = luigi.BoolParameter(
+        default=False,
+        description='If true, cache the response of successful completion checks '
+        'of tasks assigned to a worker. This can especially speed up tasks with '
+        'dynamic dependencies but assumes that the completion status does not change '
+        'after it was true the first time.',
+    )
+    skip_if_completed_pre_run: bool = ExplicitBoolParameter(
```
hiro-o918, line 378: this parameter is an alternative one of `complete_check_at_run`
Rename the parameter because luigi has very similar name parameter `check_complete_on_run` to this

hirosassa (approved):
Looks good!

kitagry:
Thank you for the great job 👍 

Do you think about compatibility for the check_complete_on_run?

PR #411, opened 2024-11-27
feat: add serializable parameter

hiro-o918:
## What

* add an object parameter implementing `gokart_serialize` and `gokart_desirialize` function

## Why

* currently we need to list all the parameter to a task
    * luigi.Parameters always flattend on the task
    * luigi.Config also cannot be passed as a config object
      * we need inherit it and use as flatten parameters
* make a set of parameters are portable to be passed to a task

## For example

The following example shows that this feature makes passing a set of parameters among tasks

```python
@dataclass(frozen=True)
class Config:
    foo: int
    bar: str

    def gokart_serialize(self) -> str:
        # dict is ordered in Python 3.7+
        return json.dumps(asdict(self))

    @classmethod
    def gokart_deserialize(cls, s: str) -> 'Config':
        return cls(**json.loads(s))


class FooTask(TaskOnKart):
    config: Config = SerializableParameter(object_type=Config)

    def run(self):
        self.dump(self.config)

class BarTask(TaskOnKart):
    config: Config = SerializableParameter(object_type=Config)

    def run(self):
        self.dump(self.config)

class Pipeline(TaskOnKart):
    config: Config = SerializableParameter(object_type=Config)
    def requires(self):
        return dict(foo=FooTask(config=self.config), bar=barTask(config=self.config),
```

### DIFF
--- a/gokart/__init__.py
+++ b/gokart/__init__.py
@@ -1,7 +1,7 @@
 from gokart.build import WorkerSchedulerFactory, build  # noqa:F401
 from gokart.info import make_tree_info, tree_info  # noqa:F401
 from gokart.pandas_type_config import PandasTypeConfig  # noqa:F401
-from gokart.parameter import ExplicitBoolParameter, ListTaskInstanceParameter, TaskInstanceParameter  # noqa:F401
+from gokart.parameter import ExplicitBoolParameter, ListTaskInstanceParameter, SerializableParameter, TaskInstanceParameter  # noqa:F401
 from gokart.run import run  # noqa:F401
 from gokart.task import TaskOnKart  # noqa:F401
 from gokart.testing import test_run  # noqa:F401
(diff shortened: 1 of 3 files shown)


kitagry (approved):
LGTM

yokomotod:
@hiro-o918 lgtm!

yokomotod (approved):
[memo]

I thought we may want more specific `Parameter` for json encode/decod-able objects, then we don't need to make custom wrapper class.

... -> we can make later if we really need it fine

Then, the name of `SerializableParameter` won't conflict / make confuse ?

... -> seems fine, we can name `JsonSerializableParameter` or something else

hiro-o918:
@yokomotod 
Nice comment.
As you mentioned,`SerializableParameter` (this PR) is a generic purpose one.
If we need more specific parameter, we can chose a detailed name.

PR #419, opened 2024-12-21
docs: add an example to show how to use type-annotations on gokart

hiro-o918:
We introduce type-annotations for gokart pipelines.
Add an example how to write them on README


hirosassa (approved):
👍

kitagry (approved):

PR #420, opened 2024-12-22
changed skip_if_completed_pre_run->task_completion_check_at_run

mski-iksm:
I've changed the parameter name and description, because it may be little confusing

### DIFF
--- a/gokart/worker.py
+++ b/gokart/worker.py
@@ -126,7 +126,7 @@ class TaskProcess(multiprocessing.Process):
         check_unfulfilled_deps: bool = True,
         check_complete_on_run: bool = False,
         task_completion_cache: Optional[Dict[str, Any]] = None,
-        skip_if_completed_pre_run: bool = True,
+        task_completion_check_at_run: bool = True,
     ) -> None:
         super(TaskProcess, self).__init__()
         self.task = task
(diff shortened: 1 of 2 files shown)


hirosassa (approved):
👍

hiro-o918:
@mski-iksm 
`task_completion_check_at_run` is confused with `check_complete_on_run` of luigi parameter.
So I renamed `task_completion_check_at_run` as  `skip_if_completed_pre_run` because `task_completion_check_at_run` will be removed from `gokart.TaskOnKart` after migrating to gokart worker.

I suppose that `task_completion_check_at_run` should be more distinguishable from `check_complete_on_run`

hiro-o918 (approved):
LGTM

mski-iksm:
@hiro-o918 

> @mski-iksm `task_completion_check_at_run` is confused with `check_complete_on_run` of luigi parameter. So I renamed `task_completion_check_at_run` as `skip_if_completed_pre_run` because `task_completion_check_at_run` will be removed from `gokart.TaskOnKart` after migrating to gokart worker.
> 
> I suppose that `task_completion_check_at_run` should be more distinguishable from `check_complete_on_run`

Thank you for the comment.

It is true that task_completion_check_at_run is similar to check_complete_on_run, but I think it is the latter that is the misleading name.
check_complete_on_run is actually “tasks will also verify that their outputs exist when they finish running, and will fail immediately if the outputs are missing”. And I think task_completion_check_at_run is a more straightforward name since it does a complete_check at run time.

Therefore, I am thinking of leaving it as task_completion_check_at_run.

PR #423, opened 2025-01-09
Log task ID to locate its cache

mamo3gr:
## Problem

We often need the task's output (cache) to examine the failed workload. Gokart (Luigi) logs the processing status of each task but we usually search for its output by listing cloud storage objects and finding the corresponding created/updated time.

## Solution

Log the task ID, which is determined by parameters and used to name its cache file.

## Discussion

As far as I know, `__repr__` and `__str__` of `gokart.TaskOnKart` seem to be used only in logging and it does not matter to change their format. I appreciate your double-checking.

### DIFF
--- a/gokart/task.py
+++ b/gokart/task.py
@@ -540,13 +540,15 @@ If you want to specify `required_columns` and `drop_columns`, please extract the
 
     def __repr__(self):
         """
-        Build a task representation like `MyTask(param1=1.5, param2='5', data_task=DataTask(id=35tyi))`
+        Build a task representation like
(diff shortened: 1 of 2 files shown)


mamo3gr:
>seem to be used only in logging

I missed the test and will add it.

gokart/task.py
```diff
@@ -540,13 +540,15 @@ def _get_module_versions(self) -> str:
 
     def __repr__(self):
         """
-        Build a task representation like `MyTask(param1=1.5, param2='5', data_task=DataTask(id=35tyi))`
+        Build a task representation like
+        `MyTask[aca2f28555dadd0f1e3dee3d4b973651](param1=1.5, param2='5', data_task=DataTask(c1f5d06aa580c5761c55bd83b18b0b4e))`
```
Hi-king, line 544: 👍

Hi-king (approved):
Thank you for the good work on improving usability!

PR #424, opened 2025-01-11
Add Option to Output All TaskOnKart Cache Hash and  File Paths in gokart.build

Hi-king:
## Summary:

To address the common issue of identifying output locations of tasks, particularly when tasks fail and users wish to restart from the last cached output, this PR introduces an option in gokart.build to output all TaskOnKart cache file paths in an organized table or tree format. This builds on last week's update (https://github.com/m3dev/gokart/pull/423), which made task hash IDs available, by providing a structured view of all task outputs.

## Changes:

Added a new parameter task_dump_config to gokart.build, which allows users to specify the output format (table or tree) and output type (print or save).
Leveraged existing functionality for generating output trees and tables of task details, making it easily accessible via build arguments.

## Sample

```python
class HelloWorldTask(gokart.TaskOnKart):
    _cache_key_date = luigi.DateMinuteParameter(default=datetime.datetime.now())
    
    def run(self):
        self.dump("Hello, World!")

class RecursiveTask(gokart.TaskOnKart):
    dep = gokart.TaskInstanceParameter()
    _cache_key_date = luigi.DateMinuteParameter(default=datetime.datetime.now())

    def requires(self):
        return self.dep

    def run(self):
        self.dump(self.load())

gokart.build(
    RecursiveTask(dep=RecursiveTask(dep=HelloWorldTask())),
    task_dump_config=TaskDumpConfig(
        mode=TaskDumpMode.TABLE, 
        output_type=TaskDumpOutputType.PRINT
    ),
    log_level=logging.INFO
)
```
Output Example:

```tsv
name	unique_id	output_paths	params	processing_time	is_complete	task_log	requires
RecursiveTask	567bdf2035c845fae3e1a629c82637bf	['./resources/__main__/RecursiveTask_567bdf2035c845fae3e1a629c82637bf.pkl']	{'dep': 'RecursiveTask-c006250dda2797f143702f5a8084fca7', '_cache_key_date': '2025-01-11T1812'}	unknown	PENDING	{}	RequiredTask(name='RecursiveTask', unique_id='c006250dda2797f143702f5a8084fca7')
[...]

PR #425, opened 2025-01-11
add python 3.13 support

Hi-king:
SSIA

I would like to use match statement (see https://github.com/m3dev/gokart/pull/424)


hirosassa (approved):
I agree with this change, but need announcement to the users because Python 3.9 is under support until this October.

Hi-king:
@hirosassa
I've managed to make the match statement work alongside Python 3.9 (https://github.com/m3dev/gokart/pull/424). 

For now, we'll just add support for Python 3.13. However, we should consider phasing out Python 3.9 soon.

mski-iksm:
LGTM!

Hi-king:
Please merge this if it's OK :)

PR #384, opened 2024-07-19
feat: add mypy plugin

hiro-o918:
## What
* implement mypy plugin for `gokart.TaskOnKart`
  *  This plugin checks the following when all the parameters are typed:
    * The parameters passed by the constructor are valid type
    * There are no missing parameters

image

PR #428, opened 2025-01-31
feat: add mypy plugin options to handle missing paramters for a task

hiro-o918:
## Background

In this [discussion](https://github.com/m3dev/gokart/pull/384#discussion_r1685638052), I initially omitted handling errors for missing parameters in the task constructor because luigi.Config allows implicit parameters, making it difficult to validate them on linting

However, I occasionally forget to pass required parameters to tasks, leading to runtime errors when executing the pipeline.

## What

This PR introduces an option for gokart mypy to raise an error when a task constructor is called without all required parameters.

Users can configure mypy to allow implicit parameter passing if needed in their project.

### DIFF
--- a/gokart/mypy.py
+++ b/gokart/mypy.py
@@ -7,13 +7,19 @@ https://github.com/python/mypy/blob/0753e2a82dad35034e000609b6e8daa37238bfaa/myp
 from __future__ import annotations
 
 import re
+import sys
+import warnings
 from collections.abc import Iterator
-from typing import Callable, Final, Literal
+from dataclasses import dataclass
+from enum import Enum
+from typing import Any, Callable, Final, Literal
 
 import luigi
 from mypy.expandtype import expand_type
 from mypy.nodes import (
+    ARG_NAMED,
     ARG_NAMED_OPT,
+    ArgKind,
     Argument,
     AssignmentStmt,
     Block,
@@ -32,6 +38,7 @@ from mypy.nodes import (
     TypeInfo,
     Var,
 )
+from mypy.options import Options
 from mypy.plugin import ClassDefContext, FunctionContext, Plugin, SemanticAnalyzerPluginInterface
 from mypy.plugins.common import (
     add_method_to_class,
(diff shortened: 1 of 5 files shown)


hirosassa (approved):
LGTM!

test/config/pyproject_error_on_missing_parameters.toml
```diff
@@ -0,0 +1,9 @@
+[tool.mypy]
+plugins = ["gokart.mypy"]
+
+[[tool.mypy.overrides]]
+ignore_missing_imports = true
+module = ["pandas.*", "apscheduler.*", "dill.*", "boto3.*", "testfixtures.*", "luigi.*"]
+
+[tool.gokart-mypy]
+error_on_missing_parameters = true
```
kitagry, line 9: I think this name is a little long. And the name of mypy is something like `disallow-` or `ignore-`. Therefore, I think `disallow_missing_parameters` is preferable. What do you think?

gokart/mypy.py
```diff
@@ -55,7 +61,50 @@
...
+
+        # TODO: support other configuration file formats if necessary.
+        if not config_file.endswith('.toml'):
+            warnings.warn('gokart mypy plugin can be configured by pyproject.toml', stacklevel=1)
+            return cls()
+
+        config = cls._parse_toml(config_file)
+        gokart_plugin_config = config.get('tool', {}).get('gokart-mypy', {})
+
+        error_on_missing_parameters = gokart_plugin_config.get('error_on_missing_parameters', False)
```
kitagry, line 97: I think parameter name should be FINAL variable.

kitagry (changes requested):

gokart/mypy.py
hiro-o918: I introduce Enum for parameters

test/config/pyproject_error_on_missing_parameters.toml
hiro-o918: updated the param name

kitagry (approved):
@hiro-o918 Sorry for the late review. Could you add documentation?

hiro-o918:
Sorry for late response, I added docs
[ddf8e6a](https://github.com/m3dev/gokart/pull/428/commits/ddf8e6a9eb15cf40c99a438bde7348fb8c6b4b88)

kitagry:
Thank you!!

PR #429, opened 2025-02-02
migration to `uv`

hirosassa:
## What is this PR

SSIA
I migrated to `uv` from `poetry` for better developer experience.


tox.ini
```diff
@@ -1,31 +1,19 @@
...
 [testenv]
-allowlist_externals = coverage
-skip_install = true
+runner = uv-venv-lock-runner
+dependency_groups = test
 commands =
-    coverage run
-    coverage xml
-    coverage report
+  {envpython} -m pytest -n auto --cov=gokart --cov-report=xml test -vv
```
hirosassa, line 9: just simplified

pyproject.toml
```diff
@@ -1,58 +1,94 @@
...
+  "/README.md",
+  "/examples",
+  "/gokart",
+  "/test",
+]
 
 [tool.ruff]
 line-length = 160
-exclude = ["venv/*", "tox/*"]
+exclude = ["venv/*", "tox/*", "examples/*"]
```
hirosassa, line 91: From v0.6 of ruff supports jupyter notebook files. Currently ignored

pyproject.toml
```diff
@@ -1,58 +1,94 @@
...
+test = [
+  "coverage",
+  "fakeredis",
+  "lupa",
+  "matplotlib",
+  "moto",
+  "mypy",
+  "pytest",
+  "pytest-cov",
+  "pytest-xdist",
```
hirosassa, line 58: added for faster testing

hiro-o918:
LGTM!

@kitagry @Hi-king 
We must update repo settings to update required CI pipelines because of changes about environment matrixes.

hiro-o918 (approved):

.github/workflows/format.yml
```diff
@@ -14,14 +14,14 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
-    - uses: actions/setup-python@v5
+    - name: Set up the latest version of uv
+      uses: astral-sh/setup-uv@v5
       with:
-        python-version: 3.12
+        enable-cache: true
+        cache-dependency-glob: "pyproject.toml"
```
yokomotod, line 21: [must?] This should be 

```suggestion
        cache-dependency-glob: "uv.lock"
```

?

now it seems to use old cache even if we've re-lock.

----

in `.github/workflows/publish.yml`, `.github/workflows/test.yml` as well.

----

[astral-sh/setup-uv@v5
](https://github.com/astral-sh/setup-uv?tab=readme-ov-file#cache-dependency-glob)

> The default is

```
cache-dependency-glob: |
  **/requirements*.txt
  **/uv.lock
```

it seems fine even if we don't set `cache-dependency-glob:` explicitly

pyproject.toml
```diff
@@ -1,58 +1,94 @@
...
 license = "MIT"
 readme = "README.md"
-homepage = "https://github.com/m3dev/gokart"
-repository = "https://github.com/m3dev/gokart"
-documentation = "https://gokart.readthedocs.io/en/latest/"
+requires-python = ">=3.9, <4"
+dependencies = [
+  "luigi",
+  "boto3",
+  "slack-sdk>=3",
```
yokomotod, line 17: [nits]

https://github.com/m3dev/gokart/pull/429/files?diff=split&w=0#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L22

`>=3` is differennt from `"^3"`, now it allows 4.x and greater. 

I think it doesn't matter, so just

```suggestion
  "slack-sdk",
```

---

`backoff` too ?

yokomotod (commented):
@hirosassa I love `uv` !

yokomotod:
[question]

`ruff` format changes just caused by ruff's version up, not uv migration ?

tox.ini
kitagry, line 9: I think it is better to write `-n auto` at  `pytest.ini_options` in pyproject.toml.

.github/workflows/format.yml
hirosassa: > it seems fine even if we don't set cache-dependency-glob: explicitly
Oh, I didn't know that. I agree with this idea 👍

pyproject.toml
hirosassa: backoff removed in [df69966](https://github.com/m3dev/gokart/pull/429/commits/df6996682e24e38cdd7f644da7a67856f7e9adb4)

.github/workflows/format.yml
hirosassa: fixed in [349cded](https://github.com/m3dev/gokart/pull/429/commits/349cdedd9e0bed4db1c0276a97526bfd1c2e99cd)

tox.ini
hirosassa: fixed in [41af399](https://github.com/m3dev/gokart/pull/429/commits/41af3999db3881094dec16e1f6a8b28eb378ecc6)

hirosassa:
@hiro-o918 
> We must update repo settings to update required CI pipelines
I updated by myself 👍

hirosassa:
@yokomotod > ruff format changes just caused by ruff's version up, not uv migration ?
Yes, it is. I should separate this kind of change, but the changes on this PR will be more complex, so I'd like to include them.

kitagry (approved):
LGTM

hirosassa:
@yokomotod could you review this PR?

yokomotod (approved):
@hirosassa lgtm!

Issue #430, opened 2025-02-05
test_build_parallel fails on macOS with spawn multiprocessing mode


inakam:
### Problem
When running parallel tasks (test_build_parallel) on MacOS using multiprocessing, the following error occurs:

```python
_pickle.PicklingError: Can't pickle <functools._lru_cache_wrapper object at 0x1303fc5c0>: 
it's not the same object as luigi.task.Task.to_str_params
```

This happens (maybe) because:
1. MacOS uses 'spawn' as the default multiprocessing start method instead of 'fork'
2. We're using `functools.lru_cache` to decorate `to_str_params` method on Task instances
3. The decorated method cannot be pickled properly when using spawn mode

### Reproduction (test_build_parallel)
```python
class _ParallelRunner(gokart.TaskOnKart[str]):
    def requires(self):
        return [_DummyTask(param=str(i)) for i in range(10)]

    def run(self):
        self.dump('done')

# The error occurs when running:
gokart.build(_ParallelRunner(), workers=20)
```

### Workarounds
1. **Switch to fork mode**: The error disappears when using fork mode on MacOS:
```python
if sys.platform == 'darwin':
    multiprocessing.set_start_method('fork')
```

2. **Cache task outputs**: The error also disappears when task outputs are cached under `resources/test/`. After the test passes by the 1st workaround and then reverting the implementation, the issue does not reproduce when re-running the test. This suggests the issue may be related to task state serialization.

### Possible Solutions
1. Use instance-based caching instead of `lru_cache`
2. Force 'fork' mode on MacOS
3. Implement a picklable cache decorator
4. Improve task output caching mechanism

inakam:
fixed
[feat: set multiprocessing start method to 'fork' for compatibility #431](https://github.com/m3dev/gokart/pull/431)
[add: CI on MacOS #433](https://github.com/m3dev/gokart/pull/433)

PR #431, opened 2025-02-05
feat: set multiprocessing start method to 'fork' for compatibility

inakam:
related issue: https://github.com/m3dev/gokart/issues/430

Setting the multiprocessing start method to 'fork'.

On MacOS, multiprocessing runs in spawn mode. 
This Pull Request is to resolve the issue of test_build_parallel failing due to this behavior. 

On Linux, fork mode is the default, so this fix is unlikely to affect its operation.(and gokart runs on Linux in most cases.)


hirosassa (approved):
LGTM

hirosassa:
https://github.com/m3dev/gokart/pull/433 needs this change. I'll merge this.

PR #432, opened 2025-02-05
feat: add an in-memory feature runnable with a single thread

utotsubasa:


gokart/in_memory/repository.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def set_value(self, id: str, obj: Any) -> None:
+        data = InMemoryData.create_data(obj)
+        self._cache[id] = data
+
+    def has(self, id: str) -> bool:
+        return id in self._cache
+
+    def remove(self, id: str) -> None:
+        assert self.has(id)
```
ujiuji1259, line 32: It would be better to provide a error messages

ujiuji1259:
Could you add documentation on how to use InMemoryTarget? I think [Additional Feature](https://github.com/m3dev/gokart/blob/4319ea505ea85829e8bce52d90172fb9cad83bff/docs/task_on_kart.rst#advanced-Features) section would be good

utotsubasa:
> Could you add documentation on how to use InMemoryTarget? 

Currently we don't have a feature to integrate `InMemoryTarget` into `TaskOnKart`, but I plan to add it.
I will document how to use it afterward!

hirosassa (commented):
added nits comment

gokart/in_memory/target.py
```diff
@@ -0,0 +1,46 @@
...
+            raise ValueError(f'No object(s) which id is {self._data_key} are stored before.')
+        time = _repository.get_last_modification_time(self._data_key)
+        return time
+
+    def _path(self) -> str:
+        # TODO: this module name `_path` migit not be appropriate
+        return self._data_key
+
+
+def make_inmemory_target(target_key: str, task_lock_params: TaskLockParams):
```
ujiuji1259, line 45: > Currently we don't have a feature to integrate InMemoryTarget into TaskOnKart, but I plan to add it.

Sounds nice! 

But, what do you think about adding the following docs in this pr? I think InMemoryTarget can be used even now by overriding `output` (and it's common way to use the custom targets/processor in gokart).

```python
def output(self):
    unique_id = self.make_unique_id() if use_unique_id else None
    # TaskLock is not supported in InMemoryTarget, so it's dummy
    task_lock_params = make_task_lock_params(
        file_path='dummy_path',
        unique_id=unique_id,
        redis_host=self.redis_host,
        redis_port=self.redis_port,
        redis_timeout=self.redis_timeout,
        raise_task_lock_exception_on_collision=False,
    )
    return make_inmemory_target(unique_id, task_lock_params)
```

After adding the interface to TaskOnKart, this docs can be fixed to use it.

ujiuji1259: Thank you for adding! Could you add the comment that the InMemoryTarget is "experimental" feature?

ujiuji1259:
added nits comment, but lgtm

test/in_memory/test_repository.py
```diff
@@ -0,0 +1,63 @@
...
+import pytest
+
+from gokart.in_memory import InMemoryCacheRepository as Repo
+
+dummy_num = 100
+
+
+class TestInMemoryCacheRepository:
+    @pytest.fixture
+    def repo(self):
```
hiro-o918, line 12:
```suggestion
    def repo(self) -> Repo:
```

gokart/in_memory/data.py
hirosassa, line 1: naming rule should be consistent
- directory name: `in_memory`
- variable name: `inmemory`

hiro-o918 (approved):
LGTM!

ujiuji1259:
LGTM

gokart/in_memory/target.py
```diff
@@ -0,0 +1,46 @@
...
+from gokart.in_memory.repository import InMemoryCacheRepository
+from gokart.target import TargetOnKart, TaskLockParams
+
+_repository = InMemoryCacheRepository()
+
+
+class InMemoryTarget(TargetOnKart):
+    def __init__(self, data_key: str, task_lock_param: TaskLockParams):
+        if task_lock_param.should_task_lock:
+            warning(f'Redis in {self.__class__.__name__} is not supported now.')
```
hiro-o918, line 14: How about raising exception instead of just showing warning?

```suggestion
            raise ValueError('Redis in {self.__class__.__name__} is not supported now.')
```

hiro-o918 (approved):

gokart/in_memory/data.py
hirosassa: @utotsubasa Could you check this comment?

hirosassa (changes requested):

gokart/in_memory/data.py
utotsubasa: I renamed all `inmemory` in variable names to `in_memory`. please check this commit!
https://github.com/m3dev/gokart/pull/432/commits/769404a50133ba80ceab87199cc9f0fba43e1403

hirosassa (approved):
LGTM!

PR #433, opened 2025-02-05
add: CI on MacOS

inakam:
related issue: https://github.com/m3dev/gokart/issues/430

While testing on MacOS is necessary, it's more expensive than Linux testing in GitHub Actions (MacOS runners consumes 10x more minutes). 
Furthermore, MacOS is rarely used in production environments, so MacOS testing is merely to verify correct functionality in local development environments. 
Therefore, I think it's sufficient to only testing for the latest Python version.

The configuration is written in a somewhat hard-to-read manner to achieve the goal of performing minimal testing on MacOS.

Desired (but unsupported) syntax:
```
matrix:
  x: [1, 2, 3, 4]
  y: [a, b, c]
  exclude:
    - x: [1, 3]
      y: [a, c]
```

Supported syntax:
```
matrix:
  x: [1, 2, 3, 4]
  y: [a, b, c]
  exclude:
    - x: 1
      y: a
    - x: 1
      y: c
    - x: 3
      y: a
    - x: 3
      y: c
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
       max-parallel: 5
       matrix:
         platform: [ubuntu-latest]
-        tox-env: ["py39", "py310", "py311", "py312", "py313"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
 
     steps:
     - uses: actions/checkout@v4
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,6 +22,7 @@ jobs:
             tox-env: "py312"
           - platform: ubuntu-latest
             tox-env: "py313"
+          # test only on latest python for macos
           - platform: macos-13
             tox-env: "py313"
           - platform: macos-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ jobs:
   tests:
     runs-on: ${{ matrix.platform }}
     strategy:
-      max-parallel: 5
+      max-parallel: 7
       matrix:
         include:
           - platform: ubuntu-latest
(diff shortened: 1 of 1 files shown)


yokomotod (changes requested):
@inakam 

you don't need another job, see #434

kitagry:
@inakam Could you rebase master?

hirosassa:
related issue on acttions
https://github.com/actions/runner/issues/857

.github/workflows/test.yml
```diff
@@ -11,8 +11,28 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        platform: [ubuntu-latest]
+        platform:
```
hirosassa, line 14: How do you think the explicit expression like following:

```suggestion
        include:
          - platform: ubuntu-latest
             python-version: "3.9"
          - platform: ubuntu-latest
             python-version: "3.10"
             ...
           - platform: macos-latest
             python-version: "3.13"
```

inakam: I respected the original writing style, but this one seems better!

hirosassa:
@inakam Could you check the result of `macos-latest, py313` run?

.github/workflows/test.yml
```diff
@@ -9,11 +9,23 @@ jobs:
...
+          - platform: ubuntu-latest
+            tox-env: "py311"
+          - platform: ubuntu-latest
+            tox-env: "py312"
+          - platform: ubuntu-latest
+            tox-env: "py313"
+          - platform: macos-13
+            tox-env: "py313"
+          - platform: macos-latest
+            tox-env: "py313"
```
hirosassa, line 28: [nits] Could you add a comment like below?
```suggestion
          # test only on latest python for macos
          - platform: macos-13
            tox-env: "py313"
          - platform: macos-latest
            tox-env: "py313"
```

hirosassa (approved):
A comment added but looks great 👍

.github/workflows/test.yml
inakam: OK! I have inserted the comment!

kitagry (approved):
LGTM

PR #434, opened 2025-02-05
CI: macos runner

yokomotod:
Sadly we have some problems on mac. (ex #430 ) Let's detect it

https://github.com/m3dev/gokart/actions/runs/13151125722

image


inakam:
issue: https://github.com/m3dev/gokart/pull/431

yokomotod:
dup #434

Issue #436, opened 2025-02-10
pytest: test results are flaky


utotsubasa:
### Summary
Pytest tests on CI are flaky and fail randomly.

Example:
https://github.com/m3dev/gokart/actions/runs/13232881004/job/36932852531

<img width="1097" alt="Image" src="https://github.com/user-attachments/assets/6963c3b9-40d7-46bd-ad85-c3a36664db01" />

hirosassa:
The detailed logs
```
=================================== FAILURES ===================================
___________________ LocalTargetTest.test_save_pandas_series ____________________
[gw0] linux -- Python 3.12.9 /home/runner/work/gokart/gokart/.tox/py312/bin/python

self = <test.test_target.LocalTargetTest testMethod=test_save_pandas_series>

    def test_save_pandas_series(self):
        obj = pd.Series(data=[1, 2], name='column_name')
        file_path = os.path.join(_get_temporary_directory(), 'test.csv')
    
        target = make_target(file_path=file_path, unique_id=None)
>       target.dump(obj)

test/test_target.py:149: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
gokart/target.py:32: in dump
    wrap_dump_with_lock(func=self._dump, task_lock_params=self._get_task_lock_params(), exist_check=self.exists)(obj)
gokart/target.py:97: in _dump
    with self._target.open('w') as f:
.tox/py312/lib/python3.12/site-packages/luigi/format.py:405: in __exit__
    self._stream.__exit__(*args)
.tox/py312/lib/python3.12/site-packages/luigi/target.py:351: in __exit__
    return super(AtomicLocalFile, self).__exit__(exc_type, exc, traceback)
.tox/py312/lib/python3.12/site-packages/luigi/target.py:331: in close
    self.move_to_final_destination()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <atomic_file name='/home/runner/work/gokart/gokart/temporary/test.csv-luigi-tmp-8033466393'>

    def move_to_final_destination(self):
>       os.replace(self.tmp_path, self.path)
[...]

PR #438, opened 2025-02-11
Fix flakiness on tests

hirosassa:
# What is this PR

fixes https://github.com/m3dev/gokart/issues/436

In this PR, I fixed flaky behavior of some test cases.

The cause of flakiness is that some tests handle the same directory simultaneously by the concurrent test run using `pytest-xdist` because such tests create temporary directory by using `_get_temporary_directory` function , but the directory name is fixed.
I fixed `_get_temporary_directroy` implementation and call it in `setUp` method to call it for each test runs.

### DIFF
--- a/test/util.py
+++ b/test/util.py
@@ -1,0 +1,8 @@
+import os
+import uuid
+
+
+# TODO: use pytest.fixture to share this functionality with other tests
+def _get_temporary_directory():
+    _uuid = str(uuid.uuid4())
+    return os.path.abspath(os.path.join(os.path.dirname(__name__), f'temporary-{_uuid}'))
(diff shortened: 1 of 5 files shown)


test/util.py
```diff
@@ -0,0 +1,8 @@
+import os
+import uuid
+
+
+# TODO: use pytest.fixture to share this functionality with other tests
```
hirosassa, line 5: In `pytest` way, this kind of function implemented as `@pytest.fixture` and use it in tests.
But the most gokart's tests are implemented in `unittest` way.
For consistency of codes, I currently implemented them in `unittest` way

hiro-o918 (approved):
LGTM!
Thank you for you fixing the issue!!

kitagry (approved):
Thank you!

PR #441, opened 2025-02-12
Feat: in memory parameter on task

utotsubasa:


gokart/task.py
```diff
@@ -134,11 +138,13 @@ def __init__(self, *args, **kwargs):
...
             self.run = wrap_run_with_lock(run_func=self.run, task_lock_params=task_lock_params)  # type: ignore
 
+        self.make_default_target = self.make_target if not self.cache_in_memory_by_default else self.make_cache_target
+
     def input(self) -> FlattenableItems[TargetOnKart]:
         return super().input()
 
     def output(self) -> FlattenableItems[TargetOnKart]:
-        return self.make_target()
+        return self.make_default_target()
```
kitagry, line 147: I prefer change here instead of overwriting a method.

```suggestion
        return self.make_target if not self.cache_in_memory_by_default()else self.make_cache_target()
```

kitagry (commented):
Sorry for the late review :pray:

PR #443, opened 2025-02-21
feat: implement `ZonedDateSecondParameter`

stakahashy:
## Summary
This PR implements `ZonedDateSecondParameter`, which supports `datetime` with `tzinfo`.

## Changes
Added `ZonedDateSecondParameterr` and its tests.


test/test_timestamp_parameter.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class TimestampParameterTest(unittest.TestCase):
+    def test_default(self):
+        with CmdlineParser.global_instance([f'{__name__}.TimestampParameterTaskWithDefault']) as cp:
+            assert cp.get_task_obj().timestamp == datetime.datetime(2025, 2, 21, 12, 0, 0, tzinfo=datetime.timezone(datetime.timedelta(hours=9)))
+
+    def test_parse_param(self):
+        with CmdlineParser.global_instance([f'{__name__}.TimestampParameterTaskWithDefault', '--timestamp', '2025-02-21T120000Z+0900']) as cp:
+            assert cp.get_task_obj().timestamp == datetime.datetime(2025, 2, 21, 12, 0, 0, tzinfo=datetime.timezone(datetime.timedelta(hours=9)))
```
kitagry, line 33: you should change time from default value in order to check it will come from command line params.

kitagry (changes requested):
Could you add some more test case? e.x) https://github.com/spotify/luigi/blob/master/test/date_parameter_test.py#L121-L136

test/test_timestamp_parameter.py
stakahashy: I fixed this and add test cases following luigi's tests.

kitagry (approved):
LGTM！

yokomotod:
great +1 for tz aware parameter!

[question]

how do you think about the naming of parameter class?

I feel that the `Timestamp` doesn't explicitly means it's "timezone aware".

by folowing the existing luigi param, for ex `DateSecondTzParamerter` (but long....

stakahashy:
> great +1 for tz aware parameter!
> 
> [question]
> 
> how do you think about the naming of parameter class?
> 
> I feel that the `Timestamp` doesn't explicitly means it's "timezone aware".
> 
> by folowing the existing luigi param, for ex `DateSecondTzParamerter` (but long....

Thank you for the comment! I think again about the naming. I now feel `ZonedDateSecondParameter` is more appropriate because its name is more consistent with luigi's naming rule and the naming indicates that timezone info is mandatory (not optional).

"Timestamp" might be a bit ambiguous as it may indicate UNIX time or some database's format.

How does it sound to you (and other reviewers?

gokart/parameter.py
```diff
@@ -119,3 +121,25 @@ def parse(self, s: str) -> S:
...
+    A TimestampParameter is a `ISO 8601 <http://en.wikipedia.org/wiki/ISO_8601>`_ formatted
+    date, time specified to the second and timezone. For example, ``2013-07-10T190738Z+0900`` specifies July 10, 2013 at
+    19:07:38 +09:00.
+    """
+
+    date_format = '%Y-%m-%dT%H%M%SZ%z'
+    _timedelta = datetime.timedelta(seconds=1)
+
+    def __init__(self, interval=1, start=None, **kwargs):
+        super().__init__(interval, start, **kwargs)
```
adieumonks, line 139: The constructor takes interval and start as arguments, but it seems they are not used anywhere. Is that correct?

yokomotod:
@stakahashy +1 for the `ZonedDateSecondParameter`.

It's similar with [Javascript's new DateTime API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/ZonedDateTime).

gokart/parameter.py
yokomotod: do we need to define `__init__()` ?

it seems fine to omit it like `DateSecondParameter`

yokomotod (changes requested):
[must] 

According to ISO 8601, the format `2013-07-10T190738Z+0900` seems wrong?

it should be

* `2013-07-10T190738Z`:  UTC
* `2013-07-10T190738+0900`: If it has the offset, no `Z`

isn't it?


https://en.wikipedia.org/wiki/ISO_8601#Coordinated_Universal_Time_(UTC)

> If the time is in [UTC](https://en.wikipedia.org/wiki/Coordinated_Universal_Time), add a Z directly after the time without a space.

gokart/parameter.py
adieumonks: The `normalize` method in `_DatetimeParameterBase` seems to normalize a given datetime by aligning it to the nearest `interval` from the `start`. However, in `TimestampParameter`, the `normalize` method is overridden, but `start` and `interval` are not used. So it might be better to either omit `__init__` entirely or use `start` and `interval` for normalization, similar to `_DatetimeParameterBase`.

stakahashy: It seems `start` and `interval` are required only to `normalize`, so as mentioned by @adieumonks , these parameters are not necessary. So I will override `__init__` as well.

stakahashy:
> [must]
> 
> According to ISO 8601, the format `2013-07-10T190738Z+0900` seems wrong?
> 
> it should be
> 
> * `2013-07-10T190738Z`:  UTC
> * `2013-07-10T190738+0900`: If it has the offset, no `Z`
> 
> isn't it?
> 
> https://en.wikipedia.org/wiki/ISO_8601#Coordinated_Universal_Time_(UTC)
> 
> > If the time is in [UTC](https://en.wikipedia.org/wiki/Coordinated_Universal_Time), add a Z directly after the time without a space.

My understanding of ISO 8601 was wrong... I will fix the format.

stakahashy:
Dear reviewers, @yokomotod @kitagry @adieumonks 

I made modifications from the last implementations including...
- the class name is renamed to `ZonedDateSecondParameter`
- inherits `luigi.Parameter` instead of  `_DateTimeParameterBase`
- Python datetime library's `isoformat` and `fromisoformat` methods are adopted instead of `strptime` and `strftime` to parse/serialize.

I'm looking for your further comments if you have any 🙏

kitagry (approved):

stakahashy:
@yokomotod @adieumonks  I suppose multiple approvals are required to merge, and if you feel comfortable, could you confirm this PR again?

adieumonks (approved):
LGTM

stakahashy:
@adieumonks Thank you!!

kitagry:
@yokomotod Could you review this? I can't merge this.

kitagry:
Sorry, this is not correct. we can bypass rule

image

Issue #124, opened 2020-02-10
Support structured logging


hirosassa:
For daily operations on public cloud environment, gokart should support structured logging.
It enables us efficient searching, easy monitoring and alert configuration on logs.

TlexCypher:
Hi, thanks for your PR! 
I just want to clarify the main goal of this PR.  
I saw a provided example, you created logging.ini for configuration of logger.
From what I understand, you want gokart to support structured logging without requiring users to manually configure `logging.ini` every time.  

To achieve this, we could implement an automatic logging setup within gokart that switches between plain text and JSON format based on an environment variable like `LOG_FORMAT=json`.  
Is my understanding correct? Or do you have any specific requirements in mind?

hirosassa:
@TlexCypher Thanks for your reply!

> an automatic logging setup within gokart that switches between plain text and JSON format based on an environment variable like LOG_FORMAT=json

This is just "what I want" feature 👍

TlexCypher:
Hi @hirosassa , thanks for your response!

I understand the feature you want to implement, but I’d like to clarify one point.

In the past five years, Python's structured logging ecosystem has evolved, and libraries like [zerolog](https://github.com/alexferl/zerolog) have emerged. If we make structured logging the default, would it be better to use something like zerolog, or should we customize the built-in logging module as shown in the example?

I haven’t run performance benchmarks yet, but if zerolog turns out to be significantly faster, I’m considering using it. What do you think?

TlexCypher:
I understand what feature you want. I work on it.

PR #444, opened 2025-02-22
Supports structured logging.

TlexCypher:
# Source issue.
https://github.com/m3dev/gokart/issues/124 (Thank you for @hirosassa )

# Summary
This PR implements SlogConfig, which supports switch functionalities structured logging and plain text logging. 
Of course, I confirmed end-to-end workflow with samples on provided issue, then structured logging and plain text logging works well.

# Changes
Added SlogConfig and its tests. 
If user export ${GOKART_LOGGING_FORMAT}=json, all logs are dumped as json, structured logging for machine.
On the other hand, the environment variables, ${GOKART_LOGGING_FORMAT}=text, all logs are dumped as plain text, for human. And also, no supported value is set, raise exception.
Also, we need user friendly interface, so implemented decorator function. So user import gokart.getLoggert instead of logging.getLogger.

# Potential Discussion Points
The interface of logging might be the important discussion point.

This is my first OSS contribution, so please let me know how my code will be improved!

### DIFF
--- a/gokart/slog_config.py
+++ b/gokart/slog_config.py
@@ -1,0 +1,63 @@
+import logging
+import os
+import sys
+from pythonjsonlogger import json
+import functools
+
+
+class SlogConfig(object):
+    """
+    LoggerConfig is for logging configuration, Utility-class.
+    This class will read an environment variable named GOKART_LOGGER_FORMAT, then switch logging configuration.
+
+    If "$GOKART_LOGGER_FORMAT=json", set logging configuration as structured logging.
+    Else if "$GOKART_LOGGER_FORMAT=text", set logging configuration as plain text, normal logging.
+    Otherwise, default setting is applied, structured logging.
+    """
+    _default_base_format = "%(message)s %(lineno)d %(pathname)s %(asctime)s %(name)s %(levelname)s %(filename)s %(lineno)s %(message)s"
+    _default_date_format = "%Y/%m/%d %H:%M:%S"
+
+    @staticmethod
+    def apply_slog_format(logger):
+        """
+        If $GOKART_LOGGER_FORMAT=json, set logging configuration as structured logging.
+        On the other hand,
+        """
+        logger_mode = os.environ.get('GOKART_LOGGER_FORMAT')
+        if not logger_mode or logger_mode.lower() == 'json':
+            if not isinstance(logger, logging.Logger):
+                return
+            if not logger.handlers:
+                handler = logging.StreamHandler(sys.stdout)
+                logger.addHandler(handler)
+
+            for handler in logger.handlers:
+                if isinstance(handler, logging.StreamHandler):
+                    fmt = handler.formatter._fmt if handler.formatter and handler.formatter._fmt else SlogConfig._default_base_format
+                    date_fmt = handler.formatter.datefmt if handler.formatter and handler.formatter.datefmt else SlogConfig._default_date_format
+
+                    formatter = json.JsonFormatter(
+                        fmt=fmt,
+                        datefmt=date_fmt,
+                    )
+                    handler.setFormatter(formatter)
+            return logger
+        # plain text mode, so nothing is applied.
+        elif logger_mode.lower() == 'text':
+            if not logger.handlers:
+                handler = logging.StreamHandler(sys.stdout)
+                logger.addHandler(handler)
+            return logger
+        else:
+            raise Exception(f"Unknown logger format: {logger_mode}")
+
+# This is the decorator method for logging.getLogger.
+def getLogger_decorator(func):
+    @functools.wraps(func)
+    def wrapper(name):
+        logger = func(name)
+        logger = SlogConfig.apply_slog_format(logger)
+        return logger
+    return wrapper
+
+
--- a/gokart/slog_config.py
+++ b/gokart/slog_config.py
@@ -7,7 +7,7 @@ from pythonjsonlogger import json
 
 class SlogConfig(object):
     """
-    LoggerConfig is for logging configuration, Utility-class.
+    SlogConfig is for logging configuration, Utility-class.
     This class will read an environment variable named GOKART_LOGGER_FORMAT, then switch logging configuration.
 
     If "$GOKART_LOGGER_FORMAT=json", set logging configuration as structured logging.
@@ -19,13 +19,18 @@ class SlogConfig(object):
     default_date_format = '%Y/%m/%d %H:%M:%S'
 
     @staticmethod
-    def apply_slog_format(logger):
+    def switch_log_format(logger):
         """
         If $GOKART_LOGGER_FORMAT=json, set logging configuration as structured logging.
         On the other hand, set logging configuration as plain text.
         """
         logger_mode = os.environ.get('GOKART_LOGGER_FORMAT')
-        if not logger_mode or logger_mode.lower() == 'json':
+        if logger_mode and logger_mode.lower() not in {'text', 'json'}:
+            raise Exception(f'Unknown logger format: {logger_mode}')
+        # plain text mode, so nothing is applied,
+        elif logger_mode and logger_mode.lower() == 'text':
+            return logger
+        else:
             if not isinstance(logger, logging.Logger):
                 return logger
             # If logger configuration was loaded, skip configuration.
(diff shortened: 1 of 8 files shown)


TlexCypher:
@hirosassa Thank you for reply. 
By the way, of cource, I should work on fixing CI errors as you said, but I think current implementation is not following Python standard when library provides logging configration, because of adding handlers. So, I will convert this PR to draft.
So I work on both fixing CI errors and new implementation.

TlexCypher:
# Current work
I'm working on fix CI errors.
Failure is ocurred on `TestTestFrameworkForPandasDataFrame::test_run_with_error` and `TestTestFrameworkForPandasDataFrame.test_run_with_namespace` when execute `tox -e py39`, but when I execure `tox -e py312`, all tests have passed.
And also, I create Python3.9 virtual env with uv, and run target test, tests failed on tox with python 3.9 has passed.
Do you know how to reproduce tox environment? I wanna fix it on the tox env.

# How to reproduce 
Executre following command on this branch(support-slog)
`$ tox -e py39` # [Expected] Exact two cases might fail.
`$tox -e py312` # [Expected] All testcases might pass.

hirosassa:
In my local environment, all of the `tox` tests run successfully 🤔
I'll take a look this deeply.

TlexCypher:
Hi, @hirosassa . I found the reason why CI errors are not fixed, and success to fix it.
So, I'll re-open this PR.
Please review it. 
Sorry for late response.

gokart/slog_config.py
```diff
@@ -0,0 +1,86 @@
...
+    default_date_format = '%Y/%m/%d %H:%M:%S'
+
+    @staticmethod
+    def apply_slog_format(logger):
+        """
+        If $GOKART_LOGGER_FORMAT=json, set logging configuration as structured logging.
+        On the other hand, set logging configuration as plain text.
+        """
+        logger_mode = os.environ.get('GOKART_LOGGER_FORMAT')
+        if not logger_mode or logger_mode.lower() == 'json':
```
hirosassa, line 28: To keep default behavior (especially for current users), if the `GOKART_LOGGER_FORMAT` is not set, I prefer the log format should be text.
What do you think about this idea?

pyproject.toml
```diff
@@ -26,6 +26,10 @@ dependencies = [
   "dill",
   "backoff",
   "typing-extensions>=4.11.0; python_version<'3.13'",
+  "python-json-logger>=3.2.1",
+  "notebook>=7.3.2",
```
hirosassa, line 30: I think this is unnecessary dependency. Could you please remove this?

pyproject.toml
```diff
@@ -26,6 +26,10 @@ dependencies = [
...
+  "tox>=4.24.1",
+  "tox-uv>=1.25.0",
```
hirosassa, line 32: These dependencies are unnecessary for production use (only for tests). Could you move these items to `test` dependecy-group?

gokart/slog_config.py
```diff
@@ -0,0 +1,86 @@
...
+    If "$GOKART_LOGGER_FORMAT=json", set logging configuration as structured logging.
+    Else if "$GOKART_LOGGER_FORMAT=text", set logging configuration as plain text, normal logging.
+    Otherwise, default setting is applied, structured logging.
+    """
+
+    default_base_format = '%(message)s %(lineno)d %(pathname)s %(asctime)s %(name)s %(levelname)s %(filename)s %(lineno)s %(message)s'
+    default_date_format = '%Y/%m/%d %H:%M:%S'
+
+    @staticmethod
+    def apply_slog_format(logger):
```
hirosassa, line 22: [IMO] This function selects log format and returns logger with the format. So there's the situation that the logger format is not "structured".
We should change the name of this function to such as `apply_log_format`, `switch_log_format`, etc.

gokart/slog_config.py
```diff
@@ -0,0 +1,86 @@
...
+
+                if not current_logger.parent:
+                    break
+
+                current_logger = current_logger.parent
+
+            if found_handler == 0:
+                last_resort_handler = logging.lastResort
+                if not last_resort_handler:
+                    last_resort_handler = logging.StreamHandler()
```
hirosassa, line 54:
```suggestion
                last_resort_handler = logging.lastResort if logging.lastResort is not None else logging.StreamHandler()
```

gokart/slog_config.py
```diff
@@ -0,0 +1,86 @@
...
+                    fmt=fmt,
+                    datefmt=date_fmt,
+                )
+                last_resort_handler.setFormatter(formatter)
+            return logger
+        # plain text mode, so nothing is applied.
+        elif logger_mode.lower() == 'text':
+            return logger
+        else:
+            raise Exception(f'Unknown logger format: {logger_mode}')
```
hirosassa, line 75: These `elif` and `else` block is short. Let's use "early return" on the start part of this function.

hirosassa (commented):
Thanks for the detailed investigation.

gokart/slog_config.py
TlexCypher: In my opinion, structured logging is more useful than text simple logging as metrics, so default logging format should be structured logging.
If user wanna change log format, user should change `GOKART_LOGGER_FORMAT`, or write `logging`.ini, logging configuration file.
But you bet, I can understand your opinion, for current user, it might be breaking changes, so avoid to make slog configuration default.

hiro-o918:
[IMO]

It seems that providing built-in logging functionality for first-party use may not fall within the scope of Gokart, as users are typically expected to choose their own logging libraries and formats.

However, allowing Gokart (or Luigi) to switch log formats could be quite helpful, as configuring this can be a bit challenging for users. It might be worth considering an option that simplifies this customization without enforcing a specific logging solution.

What do you think?

pyproject.toml
TlexCypher: Ah, this dependency is not necessary even for test.
I made a mistake, add tox and tox-uv...
So, remove this.

TlexCypher:
> [IMO]
> 
> It seems that providing built-in logging functionality for first-party use may not fall within the scope of Gokart, as users are typically expected to choose their own logging libraries and formats.
> 
> However, allowing Gokart (or Luigi) to switch log formats could be quite helpful, as configuring this can be a bit challenging for users. It might be worth considering an option that simplifies this customization without enforcing a specific logging solution.
> 
> What do you think?

Hmm, as u said, built-in structured logging functionalities is not scope for gokart, because gokart is just a ML pipeline library.
So, my implementation is following this thought.
If user writes logging.ini as usual, that configuration is prioritized.

And also, I think gokart should support structured logging as default, because users of gokart will handle bunch of data, logs, so needs for the way easily to gather metrics.
Many programming languages and libraries support structured logging, for example golang, at first, standard organization did't support structured logging, so zap, zerolog, many structured logging libraries were developed by community. 
As you seen in golang history, structured logging demand is growing, and especially for ML, we need the way to handle and gather bunch of parameters and logs easily.

gokart/slog_config.py
TlexCypher: Exactly. I adopt `swich_log_format` from your suggestion. thx.

pyproject.toml
TlexCypher: python-json-logger is necessary for adding JsonFormatter. But notebook is not, so removed.

gokart/slog_config.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+        logger_mode = os.environ.get('GOKART_LOGGER_FORMAT')
+        if logger_mode and logger_mode.lower() not in {'text', 'json'}:
+            raise Exception(f'Unknown logger format: {logger_mode}')
+        # plain text mode, so nothing is applied,
+        elif logger_mode and logger_mode.lower() == 'text':
+            return logger
+        else:
+            if not isinstance(logger, logging.Logger):
+                return logger
```
hirosassa, line 35: we can reduce this nested codes
```suggestion
        if logger_mode and logger_mode.lower() not in {'text', 'json'}:
            raise Exception(f'Unknown logger format: {logger_mode}')
        # plain text mode, so nothing is applied,
        if logger_mode and logger_mode.lower() == 'text':
            return logger
        if not isinstance(logger, logging.Logger):
            return logger
```

slog_example.py
hirosassa, line 1: [must] This file should be in `example/` directory.

hirosassa (commented):
added comments

TlexCypher:
Hi, reviewers, @hirosassa @kitagry @Hi-king @yokomotod @mski-iksm
In the internship session, we discuessed about the way to support structured logging to gokart, and then, as a conclusion, gokart should not provide fixed logging way as an OSS library.
But the implementation of this PR does not follow our conclusion, so I'll create another PR, just providing logging.ini, for structured logging.
So, I think this PR should be closed, how do you think all?

kitagry:
@TlexCypher

> So, I think this PR should be closed, how do you think all?

 Thank you for the comment. I agree with your proposal.

hirosassa:
I agree with the conclusion of the discussion.
Maybe we should provide the blo. tech blog article and/or the sample project.

TlexCypher:
Alright! I'll close it, and will create another PR, just including logging.ini.
Thank you all!

PR #445, opened 2025-02-27
Support functionalities to provide metadata as labels to object stored in GCS as cache.

TlexCypher:
## What does this PR do? 
This PR adds support for specifying labels when saving a task output to GCS.

## Why is this needed? 
Users of gokart can store some data like logs, results, parameter information etc. as cache in Google Cloud Storage(GCS).
But currently, gokart does not provide a way to add metadata labels when saving files to GCS.
So, it's so hard to track and filter the desired objects, data.
GCS has a useful field named custom object metadata, which can be added whatever you want as key-value.
So in this PR, support the way to provide this metadata to each object generated by gokart, easy to track and filter whatever you want.

## Checklist 
- CI is passing
- Code formatting follows project standards
- Necessary tests have been added
- Existing tests pass

### DIFF
--- a/gokart/gcs_obj_metadata_client.py
+++ b/gokart/gcs_obj_metadata_client.py
@@ -1,0 +1,121 @@
+from __future__ import annotations
+
+import copy
+from logging import getLogger
+from typing import Any, Dict, List, Optional, Tuple, Union
+
+import luigi
+from googleapiclient.model import makepatch
+
+from gokart.gcs_config import GCSConfig
+
+logger = getLogger(__name__)
+
+
+class GCSObjectMetadataClient(object):
+    """
+    This class is Utility-Class, so should not be initialized.
+    """
+
+    @staticmethod
+    def add_task_state_labels(
+        path: str,
+        params: Optional[List[Tuple[str, Any, luigi.Parameter]]] = None,
+    ) -> None:
+        # In gokart/object_storage.get_time_stamp, could find same call.
+        # _path_to_bucket_and_key is a private method, so, this might not be acceptable.
+        bucket, obj = GCSConfig().get_gcs_client()._path_to_bucket_and_key(path)
+
+        _response = GCSConfig().get_gcs_client().client.objects().get(bucket=bucket, object=obj).execute()
+        if _response is None:
+            logger.error(f'failed to get object from GCS bucket {bucket} and object {obj}.')
+            return
+
+        response: Dict[str, Any] = dict(_response)
+        original_metadata: Dict[Any, Any] = {}
+        if 'metadata' in response.keys():
+            _metadata = response.get('metadata')
+            if _metadata is not None:
+                original_metadata = dict(_metadata)
+
+        patched_metadata = GCSObjectMetadataClient._get_patched_obj_metadata(
+            copy.deepcopy(original_metadata),
+            params,
+        )
+
+        if not original_metadata == patched_metadata:
+            # If we use update api, existing object metadata are removed, so should use patch api.
+            # See the official document descriptions.
+            # [Link] https://cloud.google.com/storage/docs/viewing-editing-metadata?hl=ja#rest-set-object-metadata
+            update_response = (
+                GCSConfig()
+                .get_gcs_client()
+                .client.objects()
+                .patch(
+                    bucket=bucket,
+                    object=obj,
+                    body=makepatch({'metadata': original_metadata}, {'metadata': patched_metadata}),
+                )
+                .execute()
+            )
+
+            if update_response is None:
+                logger.error(f'failed to patch object {obj} in bucket {bucket} and object {obj}.')
+
+    @staticmethod
+    def _get_patched_obj_metadata(
+        metadata: Any,
+        params: Optional[List[Tuple[str, Any, luigi.Parameter]]] = None,
+    ) -> Union[Dict, Any]:
+        # If metadata from response when getting bucket and object information is not dictionary,
+        # something wrong might be happened, so return original metadata, no patched.
+        if not isinstance(metadata, dict):
+            logger.warning(f'metadata is not a dict: {metadata}, something wrong was happened when getting response when get bucket and object information.')
+            return metadata
+
+        # Maximum size of metadata for each object is 8KiB.
+        # [Link]: https://cloud.google.com/storage/quotas?hl=ja#objects
+        # And also, user_provided_labels should be prioritized rather than auto generated labels.
+        # So, at first, attach user_provided_labels, then add auto generated labels.
+        max_gcs_metadata_size, total_metadata_size, labels = 8 * 1024, 0, []
+        if params:
+            all_labels = GCSObjectMetadataClient._merge_with_user_provided_labels(params)
+            for label_name, label_value in all_labels:
+                size = len(str(label_name).encode('utf-8')) + len(str(label_value).encode('utf-8'))
+                if total_metadata_size + size > max_gcs_metadata_size:
+                    logger.warning(f'current metadata total size is {total_metadata_size} byte, and no more labels would be added.')
+                    break
+                total_metadata_size += size
+                labels.append((label_name, str(label_value)))
+
+        patched_metadata = dict(metadata)
+        for label_key, label_value in labels:
+            patched_metadata[label_key] = label_value
+        return patched_metadata
+
+    @staticmethod
+    def _merge_with_user_provided_labels(
+        params: List[Tuple[str, Any, luigi.Parameter]],
+    ) -> List[Tuple[Any, Any]]:
+        # luigi.Parameter.get_param_names() returns significant parameters only.
+        # [Link]: https://luigi.readthedocs.io/en/latest/_modules/luigi/task.html#Task.get_param_names
+        parameter_labels, parameter_key_set = [], set()
+
+        for param_name, param_value, param_obj in params:
+            if param_name == 'user_provided_gcs_labels':
+                continue
+            if param_value is None:
+                logger.warning(f'param_name={param_name} param_obj={param_obj} param_value is None. So skipped.')
+                continue
+            parameter_labels.append((param_name, param_value)) if param_obj.significant else None
+            parameter_key_set.add(param_name)
+        # Make user_provided override parameter when their key conflict.
+        # Log a warning message when this happens.
+        for param_name, param_value, _ in params:
+            if param_name == 'user_provided_gcs_labels':
+                user_provided_labels = dict(param_value)
+                for up_key, up_value in user_provided_labels.items():
+                    if up_key in parameter_key_set:
+                        logger.warning(f"{up_key} is already in parameter's name set. Override it's value with {up_value} from user provided labels.")
+                    parameter_labels.append((up_key, up_value))
+        return parameter_labels
--- a/gokart/gcs_obj_metadata_client.py
+++ b/gokart/gcs_obj_metadata_client.py
@@ -2,7 +2,8 @@ from __future__ import annotations
 
 import copy
 from logging import getLogger
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, Optional, Union
+from urllib.parse import urlsplit
 
 import luigi
 from googleapiclient.model import makepatch
@@ -17,14 +18,22 @@ class GCSObjectMetadataClient(object):
     This class is Utility-Class, so should not be initialized.
     """
 
+    # This is the copied method of luigi.gcs._path_to_bucket_and_key(path).
+    @staticmethod
+    def path_to_bucket_and_key(path):
+        (scheme, netloc, path, _, _) = urlsplit(path)
+        assert scheme == 'gs'
+        path_without_initial_slash = path[1:]
+        return netloc, path_without_initial_slash
+
     @staticmethod
     def add_task_state_labels(
         path: str,
-        params: Optional[List[Tuple[str, Any, luigi.Parameter]]] = None,
+        params: Optional[list[tuple[str, Any, luigi.Parameter]]] = None,
     ) -> None:
         # In gokart/object_storage.get_time_stamp, could find same call.
         # _path_to_bucket_and_key is a private method, so, this might not be acceptable.
-        bucket, obj = GCSConfig().get_gcs_client()._path_to_bucket_and_key(path)
+        bucket, obj = GCSObjectMetadataClient.path_to_bucket_and_key(path)
 
         _response = GCSConfig().get_gcs_client().client.objects().get(bucket=bucket, object=obj).execute()
         if _response is None:
@@ -43,7 +52,7 @@ class GCSObjectMetadataClient(object):
             params,
         )
 
-        if not original_metadata == patched_metadata:
+        if original_metadata != patched_metadata:
             # If we use update api, existing object metadata are removed, so should use patch api.
             # See the official document descriptions.
             # [Link] https://cloud.google.com/storage/docs/viewing-editing-metadata?hl=ja#rest-set-object-metadata
@@ -65,7 +74,7 @@ class GCSObjectMetadataClient(object):
     @staticmethod
     def _get_patched_obj_metadata(
         metadata: Any,
-        params: Optional[List[Tuple[str, Any, luigi.Parameter]]] = None,
+        params: Optional[list[tuple[str, Any, luigi.Parameter]]] = None,
     ) -> Union[Dict, Any]:
         # If metadata from response when getting bucket and object information is not dictionary,
         # something wrong might be happened, so return original metadata, no patched.
@@ -95,8 +104,8 @@ class GCSObjectMetadataClient(object):
 
     @staticmethod
     def _merge_with_user_provided_labels(
-        params: List[Tuple[str, Any, luigi.Parameter]],
-    ) -> List[Tuple[Any, Any]]:
+        params: list[tuple[str, Any, luigi.Parameter]],
+    ) -> list[tuple[Any, Any]]:
         # luigi.Parameter.get_param_names() returns significant parameters only.
         # [Link]: https://luigi.readthedocs.io/en/latest/_modules/luigi/task.html#Task.get_param_names
         parameter_labels, parameter_key_set = [], set()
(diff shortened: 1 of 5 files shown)


gokart/gcs_obj_metadata_client.py
```diff
@@ -0,0 +1,121 @@
...
+    @staticmethod
+    def _merge_with_user_provided_labels(
+        params: List[Tuple[str, Any, luigi.Parameter]],
+    ) -> List[Tuple[Any, Any]]:
+        # luigi.Parameter.get_param_names() returns significant parameters only.
+        # [Link]: https://luigi.readthedocs.io/en/latest/_modules/luigi/task.html#Task.get_param_names
+        parameter_labels, parameter_key_set = [], set()
+
+        for param_name, param_value, param_obj in params:
+            if param_name == 'user_provided_gcs_labels':
```
kitagry, line 105: Maybe, this condition is meaningless. because `user_provided_gcs_labels` is significant=False, and you omit no significant parameter in TaskOnKart.dump.

kitagry (changes requested):

gokart/gcs_obj_metadata_client.py
```diff
@@ -0,0 +1,121 @@
...
+        _response = GCSConfig().get_gcs_client().client.objects().get(bucket=bucket, object=obj).execute()
+        if _response is None:
+            logger.error(f'failed to get object from GCS bucket {bucket} and object {obj}.')
+            return
+
+        response: Dict[str, Any] = dict(_response)
+        original_metadata: Dict[Any, Any] = {}
+        if 'metadata' in response.keys():
+            _metadata = response.get('metadata')
+            if _metadata is not None:
```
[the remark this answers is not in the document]
TlexCypher: Because to use `make_patch`.
If wanna use PATCH method, original metadata might be necessary.

gokart/gcs_obj_metadata_client.py
TlexCypher: Ahh, maybe this is necessary. 
This is a little bit confused, below is the kind of interface in my mind.

```python
class SampleTask(gokart.TaskOnKart):
    sample_parameter = luigi.Parameter(default='this is sample parameter')
    parent_boolean_parameter = luigi.BoolParameter(default=True)
    user_provided_gcs_labels = luigi.DictParameter(default={'empty': True, 'created_by': 'foo', 'timestamp': 'today'})

    def run(self):
        self.dump(obj='sample output')
```

When user wanna add labels to gcs object, user should declare luigi.Parameter named user_provided_gcs_labels. And when it's declared `parameter.significant` would be True.
So this if statement is necessary.
And also, all the parameters would be passed to dump method, so metadata for each Task is gonna to be add to each cached object.

gokart/task.py
```diff
@@ -359,7 +359,12 @@ def dump(self, obj: Any, target: Union[None, str, TargetOnKart] = None) -> None:
         PandasTypeConfigMap().check(obj, task_namespace=self.task_namespace)
         if self.fail_on_empty_dump and isinstance(obj, pd.DataFrame):
             assert not obj.empty
-        self._get_output_target(target).dump(obj, lock_at_dump=self._lock_at_dump)
+
+        params = []
+        for param_name, param_obj in self.get_params():
+            params.append((param_name, self.param_kwargs[param_name], param_obj)) if param_obj.significant else None
+
+        self._get_output_target(target).dump(obj, lock_at_dump=self._lock_at_dump, params=params)
```
kitagry, line 367: Could you add testcase with Mock of TargetOnKart?

test/test_gcs_obj_metadata_client.py
```diff
@@ -0,0 +1,53 @@
+import datetime
+import unittest
+
+import luigi
+
+from gokart.gcs_obj_metadata_client import GCSObjectMetadataClient
+
+
+class TestGCSObjectMetadataClient(unittest.TestCase):
+    def test_get_patched_obj_metadata(self):
```
stakahashy, line 10: I prefer to split this test method by param like `test_get_patched_optional_parameter_metadata`, `test_get_patched_int_parameter`,... to easily check which parameter causes the error.

test/test_gcs_obj_metadata_client.py
```diff
@@ -0,0 +1,53 @@
...
+        param0 = luigi.OptionalParameter(default=None)
+        param1 = luigi.Parameter(default='param1_value')
+        param2 = luigi.IntParameter(default=100)
+        param3 = luigi.BoolParameter(default=False)
+        param4 = luigi.OptionalStrParameter(default='optional string')
+        param5 = luigi.DateParameter(default=datetime.date.today())
```
stakahashy, line 16: To minimize the risk of unexpected changes, I recommend to avoid using `today()`, instead I give the values explicitly like `datetime.date(2025, 2, 27)`.

test/test_gcs_obj_metadata_client.py
```diff
@@ -0,0 +1,53 @@
...
+        param6 = luigi.MonthParameter(default=datetime.date.today())
+        param7 = luigi.YearParameter(default=datetime.date.today())
+        param8 = luigi.DateHourParameter(default=datetime.date.today())
```
stakahashy, line 19: For the following 3 lines, you may need `datetime.datetime`, not `datetime.date`

stakahashy (commented):
I am yet to read the implementation carefully, I left some comments about your testing code.

gokart/target.py
```diff
@@ -93,9 +94,11 @@ def _load(self) -> Any:
         with self._target.open('r') as f:
             return self._processor.load(f)
 
-    def _dump(self, obj) -> None:
+    def _dump(self, obj, params: Optional[list[tuple[str, Any, luigi.Parameter]]] = None) -> None:
         with self._target.open('w') as f:
             self._processor.dump(obj, f)
+        if self.path().startswith('gs://'):
+            GCSObjectMetadataClient.add_task_state_labels(path=self.path(), params=params)
```
adieumonks, line 101: If we were to support S3, would the S3 check be implemented here?
The current implementation seems to work, so it looks good for now. However, it feels a bit odd for `SingleFileTarget` to be concerned about whether it will be saved to GCS.

gokart/gcs_obj_metadata_client.py
```diff
@@ -0,0 +1,103 @@
...
+        params: Optional[list[tuple[str, Any, luigi.Parameter]]] = None,
+    ) -> Union[Dict, Any]:
+        # If metadata from response when getting bucket and object information is not dictionary,
+        # something wrong might be happened, so return original metadata, no patched.
+        if not isinstance(metadata, dict):
+            logger.warning(f'metadata is not a dict: {metadata}, something wrong was happened when getting response when get bucket and object information.')
+            return metadata
+
+        # Maximum size of metadata for each object is 8KiB.
+        # [Link]: https://cloud.google.com/storage/quotas?hl=ja#objects
```
hirosassa, line 87: The document link should be English version. Could you check all of them?
```suggestion
        # [Link]: https://cloud.google.com/storage/quotas#objects
```

gokart/gcs_obj_metadata_client.py
```diff
@@ -0,0 +1,103 @@
...
+        metadata: Any,
+        params: Optional[list[tuple[str, Any, luigi.Parameter]]] = None,
+    ) -> Union[Dict, Any]:
+        # If metadata from response when getting bucket and object information is not dictionary,
+        # something wrong might be happened, so return original metadata, no patched.
+        if not isinstance(metadata, dict):
+            logger.warning(f'metadata is not a dict: {metadata}, something wrong was happened when getting response when get bucket and object information.')
+            return metadata
+
+        # Maximum size of metadata for each object is 8KiB.
```
hirosassa, line 86: [nits]
```suggestion
        # Maximum size of metadata for each object is 8 KiB.
```

gokart/gcs_obj_metadata_client.py
```diff
@@ -0,0 +1,103 @@
...
+                size = len(str(label_name).encode('utf-8')) + len(str(label_value).encode('utf-8'))
+                if total_metadata_size + size > max_gcs_metadata_size:
+                    logger.warning(f'current metadata total size is {total_metadata_size} byte, and no more labels would be added.')
+                    break
+                total_metadata_size += size
+                labels.append((label_name, str(label_value)))
+
+        patched_metadata = dict(metadata)
+        for label_key, label_value in labels:
+            patched_metadata[label_key] = label_value
```
hirosassa, line 102:
```suggestion
        patched_metadata = dict(metadata) | dict(labels)
```

gokart/task.py
```diff
@@ -359,7 +359,12 @@ def dump(self, obj: Any, target: Union[None, str, TargetOnKart] = None) -> None:
         PandasTypeConfigMap().check(obj, task_namespace=self.task_namespace)
         if self.fail_on_empty_dump and isinstance(obj, pd.DataFrame):
             assert not obj.empty
-        self._get_output_target(target).dump(obj, lock_at_dump=self._lock_at_dump)
+
+        params = []
+        for param_name, param_obj in self.get_params():
+            params.append((param_name, self.param_kwargs[param_name], param_obj)) if param_obj.significant else None
```
hirosassa, line 365: immutability is the matter for readability.
```suggestion
        params = [(param_name, self.param_kwargs[param_name], param_obj) for param_name, param_obj in self.get_params() if param_obj.significant else None]
```

gokart/gcs_obj_metadata_client.py
```diff
@@ -0,0 +1,103 @@
...
+logger = getLogger(__name__)
+
+
+class GCSObjectMetadataClient(object):
+    """
+    This class is Utility-Class, so should not be initialized.
+    This class used for adding metadata as labels.
+    """
+
+    # This is the copied method of luigi.gcs._path_to_bucket_and_key(path).
```
hirosassa, line 22: nice comment 👍

hirosassa (changes requested):
put some comments

gokart/target.py
TlexCypher: Thank you for great thoughtful comments!
Hmm, as you said, in the near future, gokart definitely needs to support S3.
And then, `TargetOnKart` should not know specific storage class.
Then, I come up with one way to avoid that `TargetOnKart` knows the specific storage exists, implement `ObjectMetadataClient` class, that will work as interface, abstract class, whose responsibility add task state labels to storage, and `GCSObjectMetadataClient`, `S3ObjectMetadataClient` would inherit `ObjectMetadataClient` class.
Do you think it's better?

adieumonks: Yes, I agree that creating `ObjectMetadataClient` is a good idea to make `TargetOnKart` independent of specific storage implementations.
As you said, this is something we can work on in the future, and it may not be necessary to implement it in this PR.

TlexCypher: Thank you for your work. I love this advice, then I'll create another PR, so wait for it!

gokart/task.py
TlexCypher: I'm new to mock. So I wanna confirm what you want me to add.
The reason why you wanna add mock testcase, make sure whether an entire dump workflow would be fine.
Then,  is this kind of what you expect?

```python
import unittest
from unittest.mock import patch, MagicMock
import gokart

class SampleTask(gokart.TaskOnKart):
    def output(self):
        return self.make_target('sample.pkl')

class TestGokartTask(unittest.TestCase):
    @patch.object(gokart.TargetOnKart, 'load')
    def test_mock_target_on_kart(self, mock_load):

        mock_load.return_value = {'mocked_key': 'mocked_value'}

        task = SampleTask()

        result = task.output().load()
        # check some metadata?
        self.assertEqual(result, {'mocked_key': 'mocked_value'})


if __name__ == '__main__':
    unittest.main()

```

test/test_gcs_obj_metadata_client.py
```diff
@@ -0,0 +1,447 @@
+# mypy: ignore-errors
```
TlexCypher, line 1: As I wrote, by luigi specification,  can't get parameter instance.
And then, I can't come up with an exact idea when mypy enables and all CI errors are fixed.
If you have any advice, please let me know.

gokart/gcs_obj_metadata_client.py
```diff
@@ -0,0 +1,100 @@
...
+    @staticmethod
+    def path_to_bucket_and_key(path):
```
hiro-o918, line 24: please add type annotations, though this function is copied.
```suggestion
    def path_to_bucket_and_key(path: str) -> tuple[str, str]:
```

gokart/gcs_obj_metadata_client.py
```diff
@@ -0,0 +1,100 @@
...
+
+import luigi
+from googleapiclient.model import makepatch
+
+from gokart.gcs_config import GCSConfig
+
+logger = getLogger(__name__)
+
+
+class GCSObjectMetadataClient(object):
```
hiro-o918, line 16: [nits]
```suggestion
class GCSObjectMetadataClient:
```

gokart/gcs_obj_metadata_client.py
```diff
@@ -0,0 +1,100 @@
...
+    def _get_patched_obj_metadata(
+        metadata: Any,
+        params: Optional[list[tuple[str, Any, luigi.Parameter]]] = None,
+    ) -> Union[Dict, Any]:
+        # If metadata from response when getting bucket and object information is not dictionary,
+        # something wrong might be happened, so return original metadata, no patched.
+        if not isinstance(metadata, dict):
+            logger.warning(f'metadata is not a dict: {metadata}, something wrong was happened when getting response when get bucket and object information.')
+            return metadata
+
```
hiro-o918, line 85: why don't you return metadata first, to avoid nest of `if params`
```
if not params:
    return metadata
```

gokart/task.py
```diff
@@ -359,7 +359,9 @@ def dump(self, obj: Any, target: Union[None, str, TargetOnKart] = None) -> None:
         PandasTypeConfigMap().check(obj, task_namespace=self.task_namespace)
         if self.fail_on_empty_dump and isinstance(obj, pd.DataFrame):
             assert not obj.empty
-        self._get_output_target(target).dump(obj, lock_at_dump=self._lock_at_dump)
+
+        params = [(param_name, self.param_kwargs[param_name], param_obj) for param_name, param_obj in self.get_params() if param_obj.significant]
```
hiro-o918, line 363: how about using `to_str_params` to get params for labeling?
It seems to be more clear to pass  `dict[str, str]` as labels than  `list[tuple[str, Any, luigi.Parameter]]` 

https://github.com/spotify/luigi/blob/e354f61d3195f40711954a9f6cd89148dea51123/luigi/task.py#L517-L529

hiro-o918: using `to_str_params`, we can also handle visibility of parameters to avoid save credentials as labels of GCS.

TlexCypher: Thanks for your helpful suggestion. I fully agree with you.
If I adopt this suggestion, we don't need to dismiss mypy in test_gcs_object_metadata_client.
That is great side effect.

TlexCypher: Thank you for helpful and nice suggestion.

gokart/gcs_obj_metadata_client.py
```diff
@@ -0,0 +1,100 @@
...
+class GCSObjectMetadataClient:
+    """
+    This class is Utility-Class, so should not be initialized.
+    This class used for adding metadata as labels.
+    """
+
+    # This is the copied method of luigi.gcs._path_to_bucket_and_key(path).
+    @staticmethod
+    def path_to_bucket_and_key(path: str) -> tuple[str, str]:
```
kitagry, line 23: Could you rename `_ path_to_bucket_and_key` ? I think this method is not called from other file.

gokart/target.py
```diff
@@ -27,11 +28,11 @@ def exists(self) -> bool:
     def load(self) -> Any:
         return wrap_load_with_lock(func=self._load, task_lock_params=self._get_task_lock_params())()
 
-    def dump(self, obj, lock_at_dump: bool = True) -> None:
+    def dump(self, obj, lock_at_dump: bool = True, params: Optional[dict[Any, str]] = None) -> None:
```
kitagry, line 31: I think `params` is too common name. Could you rename `task_params`?

kitagry, line 31: I think `params` is too common name. Could you rename `task_params`?

test/test_gcs_obj_metadata_client.py
kitagry: If you prefer this, you can use `pytest.mark.parameterize`.

kitagry (approved):
LGTM

TlexCypher:
@kitagry Thanks for giving me LGTM.
I'll follow your nits all comments.

gokart/task.py
```diff
@@ -359,7 +359,8 @@ def dump(self, obj: Any, target: Union[None, str, TargetOnKart] = None) -> None:
         PandasTypeConfigMap().check(obj, task_namespace=self.task_namespace)
         if self.fail_on_empty_dump and isinstance(obj, pd.DataFrame):
             assert not obj.empty
-        self._get_output_target(target).dump(obj, lock_at_dump=self._lock_at_dump)
+
+        self._get_output_target(target).dump(obj, lock_at_dump=self._lock_at_dump, task_params=super().to_str_params(only_significant=True))
```
hiro-o918, line 363: [must]
https://luigi.readthedocs.io/en/stable/api/luigi.task.html#luigi.task.Task.to_str_params

please specify only public paparameters
```suggestion
        self._get_output_target(target).dump(obj, lock_at_dump=self._lock_at_dump, task_params=super().to_str_params(only_significant=True, only_public=True))
```

gokart/target.py
```diff
@@ -135,10 +138,10 @@ def _load(self) -> Any:
         self._remove_temporary_directory()
         return model
 
-    def _dump(self, obj) -> None:
+    def _dump(self, obj, task_params: Optional[dict[Any, str]] = None) -> None:
```
hiro-o918, line 141: accding to `to_str_params` function, it will be dict[str, str]
```suggestion
    def _dump(self, obj, task_params: Optional[dict[str, str]] = None) -> None:
```

hiro-o918: https://luigi.readthedocs.io/en/stable/api/luigi.task.html#luigi.task.Task.to_str_params

TlexCypher: Ahh, thank you for suggestion.
As you showed, `to_str_params` returns `dict[str, str]`.

hiro-o918 (approved):
LGTM !!!
thank you for your great contribution!

PR #446, opened 2025-03-03
Support functionalities to add user-provided, original labels.

TlexCypher:
## Related works
https://github.com/m3dev/gokart/pull/445

By merging this PR, gokart can add task's parameter information to gcs cache.

# What does this PR do?
This PR adds support for user-provided, original labels when saving a task output to GCS.

## Why is this needed?
Users of gokart can store some data like logs, results, parameter information etc. as cache in Google Cloud Storage(GCS).
For now, gokart can provide labels, generated from each task parameter information when saving files to GCS.
But, when getting number of cached objects growing, in some cases, user want to add original labels to cache for more ease filtering.
So in this PR, support those kind way.

## Checklist
- CI is passing
- Code formatting follows project standards.
- Necessary tests have been added.
- Existing tests pass.


gokart/task.py
```diff
@@ -350,17 +350,22 @@ def _flatten_recursively(dfs):
...
     @overload
-    def dump(self, obj: T, target: None = None) -> None: ...
+    def dump(self, obj: T, target: None = None, user_provided_labels: Optional[dict[Any, Any]] = None) -> None: ...
 
     @overload
-    def dump(self, obj: Any, target: Union[str, TargetOnKart]) -> None: ...
+    def dump(self, obj: Any, target: Union[str, TargetOnKart], user_provided_labels: Optional[dict[Any, Any]] = None) -> None: ...
 
-    def dump(self, obj: Any, target: Union[None, str, TargetOnKart] = None) -> None:
+    def dump(self, obj: Any, target: Union[None, str, TargetOnKart] = None, user_provided_labels: Optional[dict[Any, Any]] = None) -> None:
```
kitagry, line 358: maybe, `user_provided_labels` name is confusing. when user call task.dump, user can't predict what the labels will be.

kitagry, line 358: I think label's key should be string.

gokart/gcs_obj_metadata_client.py
```diff
@@ -71,30 +77,63 @@ def add_task_state_labels(
...
+        total_metadata_size, labels = GCSObjectMetadataClient._add_labels_to_metadata(
+            normalized_user_provided_labels, total_metadata_size, max_gcs_metadata_size, labels, has_seen_keys
+        )
+        _, labels = GCSObjectMetadataClient._add_labels_to_metadata(
+            normalized_task_params_labels, total_metadata_size, max_gcs_metadata_size, labels, has_seen_keys
+        )
+        return dict(metadata) | dict(labels)
+
+    @staticmethod
+    def _add_labels_to_metadata(labels_dict, total_metadata_size, max_gcs_metadata_size, labels, has_seen_keys):
```
kitagry, line 124: Could you add type for args?

kitagry (changes requested):

gokart/task.py
TlexCypher: I have some candidates for this variable.
Which one is better, do you think?

1. user_labels
2. custom_labels
3. provided_labels
4. user_defnined_labels

I think 2 and 4 are better.

gokart/gcs_obj_metadata_client.py
```diff
@@ -71,30 +77,63 @@ def add_task_state_labels(
...
+    @staticmethod
+    def _normalize_labels(task_params: Optional[dict[Any, str]], user_provided_labels: Optional[dict[Any, Any]]) -> tuple[dict[Any, str], dict[Any, str]]:
+        def _normalize_labels_helper(_params: Optional[dict[Any, Any]]) -> dict[Any, str]:
+            return {str(key): str(value) for key, value in _params.items()} if _params else {}
+
+        return (
+            _normalize_labels_helper(task_params),
+            _normalize_labels_helper(user_provided_labels),
+        )
+
```
adieumonks: If `_normalize_labels` does the same thing for both `task_params` and `user_provided_labels`, it would be better to have a single argument (e.g., `labels`) and call it twice.

gokart/gcs_obj_metadata_client.py
```diff
@@ -71,30 +77,63 @@ def add_task_state_labels(
...
+        # There is a possibility that the keys of user-provided labels(user_provided_labels) may conflict with those generated from task parameters (task_params_labels).
+        # However, users who utilize user_provided_labels are no longer expected to search using the labels generated from task parameters.
+        # Instead, users are expected to search using the labels they provided.
+        # Therefore, in the event of a key conflict, the value registered by the user-provided labels will take precedence.
+        total_metadata_size, labels = GCSObjectMetadataClient._add_labels_to_metadata(
+            normalized_user_provided_labels, total_metadata_size, max_gcs_metadata_size, labels, has_seen_keys
+        )
+        _, labels = GCSObjectMetadataClient._add_labels_to_metadata(
+            normalized_task_params_labels, total_metadata_size, max_gcs_metadata_size, labels, has_seen_keys
+        )
```
adieumonks: I think `GCSObjectMetadataClient` only needs to focus on adding metadata to the cache and doesn't necessarily need to be aware of the differences between `task_params` and `user_provided_labels`

TlexCypher: Thanks for your comment. I understand that your point is that GCSObjectMetadataClient shouldn't need to distinguish between user-provided labels and labels generated from parameters. I agree with that, and one possible approach could be to merge both sets of labels beforehand as a preprocessing step. However, as mentioned in the comment, there is a possibility that the keys of user-provided labels and parameter-generated labels could conflict. Based on the use case, I want to prioritize the values of user-provided labels. If we implement this in a straightforward way, we would end up with multiple for loops performing similar processing. To avoid that redundancy, I extracted the logic into the _add_labels_to_metadata function, which led to the current implementation.

adieumonks: Would it be a bad idea to overwrite `task_params` with `user_provided_labels` before calling the function, like `GCSObjectMetadataClient.add_task_state_labels(task_params | user_provided_labels)?`

gokart/gcs_obj_metadata_client.py
```diff
@@ -71,30 +81,63 @@ def add_task_state_labels(
             if update_response is None:
                 logger.error(f'failed to patch object {obj} in bucket {bucket} and object {obj}.')
 
+    @staticmethod
+    def _normalize_labels(labels: Optional[dict[str, str]]) -> dict[str, str]:
+        def _normalize_labels_helper(_params: Optional[dict[str, Any]]) -> dict[str, str]:
```
hiro-o918, line 86: [ask]
why do you create this inner function?
I feel this is enough for this purpose.

```python
    @staticmethod
    def _normalize_labels(labels: Optional[dict[str, str]]) -> dict[str, str]:
        return {str(key): str(value) for key, value in labels.items()} if labels else {}
```

gokart/gcs_obj_metadata_client.py
```diff
@@ -71,30 +81,63 @@ def add_task_state_labels(
             if update_response is None:
                 logger.error(f'failed to patch object {obj} in bucket {bucket} and object {obj}.')
 
+    @staticmethod
+    def _normalize_labels(labels: Optional[dict[str, str]]) -> dict[str, str]:
```
hiro-o918, line 85: this annotation seems to be invalid, since `_normalize_labels` tries to convert `Optional[dict[str, Any]` to `Optional[dict[str, str]`
```suggestion
    def _normalize_labels(labels: Optional[dict[str, Any]]) -> dict[str, str]:
```

gokart/gcs_obj_metadata_client.py
```diff
@@ -71,30 +81,63 @@ def add_task_state_labels(
...
+        total_metadata_size, labels = GCSObjectMetadataClient._add_labels_to_metadata(
+            normalized_custom_labels, total_metadata_size, max_gcs_metadata_size, labels, has_seen_keys
+        )
+        _, labels = GCSObjectMetadataClient._add_labels_to_metadata(
+            normalized_task_params_labels, total_metadata_size, max_gcs_metadata_size, labels, has_seen_keys
+        )
+        return dict(metadata) | dict(labels)
+
+    @staticmethod
+    def _add_labels_to_metadata(
```
hiro-o918, line 126: hmm, mutable updating `has_seen_keys` can made some bugs.
why don't you return `has_seen_keys` from `_add_labels_to_metadata` like following?
```suggestion
    def _add_labels_to_metadata(
            labels_dict: dict[str, str], total_metadata_size: int, max_gcs_metadata_size: int, labels: Optional[list[tuple[str, str]]] = None, Optional[has_seen_keys: set[str]] = None,
    ) -> tuple[int, list[tuple[str, str]]]:
        labels = copy.copy(labels) if labels else []
        has_seen_keys = copy.copy(has_seen_keys) if has_seen_keys else set({})
        ...
        return total_metadata_size, labels, has_seen_keys
```

test/test_gcs_obj_metadata_client.py
```diff
@@ -16,35 +16,135 @@ def run(self):
...
+        got_norm_task_params_labels = GCSObjectMetadataClient._normalize_labels(None)
+        got_norm_custom_labels = GCSObjectMetadataClient._normalize_labels(None)
+        self.assertIsInstance(got_norm_task_params_labels, dict)
+        self.assertIsInstance(got_norm_custom_labels, dict)
+        self.assertEqual(got_norm_task_params_labels, {})
+        self.assertEqual(got_norm_custom_labels, {})
+
+    def test_normalize_labels_only_task_params(self):
+        got_norm_task_params_labels = GCSObjectMetadataClient._normalize_labels(self.task_params)
+        got_norm_custom_labels = GCSObjectMetadataClient._normalize_labels(None)
```
hiro-o918, line 51: [should]
we do not need to tests combinations `task_params` and `custom_labels` on this function,
because `_normalize_labels` does not aware them.

please focus on how `_normalize_labels` converts optional[dict[str, Any]] to dict[str, str].

test/test_gcs_obj_metadata_client.py
```diff
@@ -16,35 +16,135 @@ def run(self):
...
+
+        self.assertIsInstance(got_norm_task_params_labels, dict)
+        self.assertIsInstance(got_norm_custom_labels, dict)
```
hiro-o918, line 54: please assert converted values not only keys

hiro-o918: can you compare actual dict with expected dict like this.
It will be clearer to show the results

```
actual = GCSObjectMetadataClient._normalize_labels(self.task_params)
expected = {
  'param1': ...,
  'param2': ...
}

self.assertEqual(actual, expected)
```

gokart/gcs_obj_metadata_client.py
```diff
@@ -18,23 +18,32 @@ class GCSObjectMetadataClient:
     This class used for adding metadata as labels.
     """
 
+    @staticmethod
+    def _is_log_related_path(path: str) -> bool:
+        return (
+            ('log/random_seed' in path)
```
hiro-o918, line 24: how about this?

```suggestion
        return path.startwith('log/')
```

hiro-o918 (changes requested):

gokart/gcs_obj_metadata_client.py
TlexCypher: Ahh. As you said, should return has_seen_keys as same as labels.
But I have one question.
In python, set, list, dict, they are passed as reference.
But in your implementation, why do you suggest to copy it?
For immuability? I'm wondering what is the best practice in some kind of this case.

gokart/gcs_obj_metadata_client.py
TlexCypher: In my opinion, I think your suggeston is not acceptable.
Because if user wants to change default place to dump from __main__ to log/, no labels would be added.
The reason why I write as following implementation, `log/{hogehoge}` is hard coded in gokart implementation, so these passes are fixed.

test/test_gcs_obj_metadata_client.py
TlexCypher: In my opinion, this is not necessary.
Because, `_normalize_labels` iterates dictionary, which is included as arguments, and convert into string.
Expected value is depending on `__str__` method logic for each type, and we need to believe this method is well-tested enough.
So we need to write in the test script is key assertion only, no more tests we need.

gokart/gcs_obj_metadata_client.py
TlexCypher: I don't know what `GCSObjectMetadataClient.add_task_state_labels(task_params | user_provided_labels)` means, so I search it. From Python3.9, | operator merges dictionary, and right side is prioritized.
But I wanna log key conflict event, because if event is captured as log, user can know key conflicts.
What do you think?

TlexCypher: Off course, we number total keys, then after merging two dictionary, if the number of total key were changed, key conflicts might be happened, so we can log it.
But, some meaningless code would be written, to detect what exact key causes key conflicts.

test/test_gcs_obj_metadata_client.py
hiro-o918: @TlexCypher 
I understand!

gokart/gcs_obj_metadata_client.py
hiro-o918: @TlexCypher 
As you said, they are passed as reference, which also update original variables of arguments.
For example, the following assertion fails since the original `l` is also updated.

```python
def update_list(l: list[str]) -> list[str]:
    l.append('new')
    return l


l = ['old']
new_l = update_list(l)

assert len(l) < len(new_l)
```

This makes sometimes developer confused, e.g. comparing original list and new one.
The function name `_add_labels_to_metadata` does not imply to update passed arguments, but just return new 
metadata.
For many cases, I prefer make variable to be immutable, thus I suggested to copy them

gokart/gcs_obj_metadata_client.py
hiro-o918: @TlexCypher 
OK!
But I cannot find the reason to specify the subpaths, please add comments the reasons.

Also, imo, using regex is more suitable for this case like `r'^log/(processing_time|...).+\.txt$`?
Current implementation will be true for paths like`foo/log/pprocessing_time`.
Is this correct?

gokart/gcs_obj_metadata_client.py
TlexCypher: Thank you for great help.
I can clearly understand why immutability is necessary.
Thus, I accept your suggestion!

gokart/gcs_obj_metadata_client.py
TlexCypher: Using regex is better, as you said.

gokart/gcs_obj_metadata_client.py
```diff
@@ -71,30 +81,63 @@ def add_task_state_labels(
...
+        total_metadata_size, labels, has_seen_keys = GCSObjectMetadataClient._add_labels_to_metadata(
+            normalized_custom_labels, total_metadata_size, max_gcs_metadata_size
+        )
+        _, labels, _ = GCSObjectMetadataClient._add_labels_to_metadata(
+            normalized_task_params_labels, total_metadata_size, max_gcs_metadata_size, labels, has_seen_keys
+        )
+        return dict(metadata) | dict(labels)
+
+    @staticmethod
+    def _add_labels_to_metadata(
```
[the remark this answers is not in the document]
TlexCypher: I think creating a data class just for this is redundant. It would be wiser to simply rename the method.

hiro-o918 (approved):
LGTM!

gokart/gcs_obj_metadata_client.py
```diff
@@ -71,30 +75,63 @@ def add_task_state_labels(
             if update_response is None:
                 logger.error(f'failed to patch object {obj} in bucket {bucket} and object {obj}.')
 
+    @staticmethod
+    def _normalize_labels(labels: Optional[dict[str, Any]]) -> dict[str, str]:
```
yokomotod, line 79: [nits]

you can write `dict[str, Any] | None` since 3.10, or, 3.7 with `from __future__ import annotations`

gokart/gcs_obj_metadata_client.py
```diff
@@ -71,30 +75,63 @@ def add_task_state_labels(
...
+        # There is a possibility that the keys of user-provided labels(custom_labels) may conflict with those generated from task parameters (task_params_labels).
+        # However, users who utilize custom_labels are no longer expected to search using the labels generated from task parameters.
+        # Instead, users are expected to search using the labels they provided.
+        # Therefore, in the event of a key conflict, the value registered by the user-provided labels will take precedence.
+        total_metadata_size, labels, has_seen_keys = GCSObjectMetadataClient._add_labels_with_size_limitation(
+            normalized_custom_labels, total_metadata_size, max_gcs_metadata_size
+        )
+        _, labels, _ = GCSObjectMetadataClient._add_labels_with_size_limitation(
+            normalized_task_params_labels, total_metadata_size, max_gcs_metadata_size, labels, has_seen_keys
+        )
```
kitagry, line 110: I think these methods are a bit confusing. I think we can separate two phase.

1. merge `normalized_task_params_labels` and `normalized_custom_labels`
2. delete oversized labels 

I think you shouldn't multiple things in a method.

kitagry (approved):
I added nits comment. LGTM

gokart/gcs_obj_metadata_client.py
TlexCypher: Thank you for educational and helpful comment!
I'm not sure Optional expression is old expression.

hirosassa (approved):
LGTM!

yokomotod (approved):

gokart/gcs_obj_metadata_client.py
```diff
@@ -97,41 +97,49 @@ def _get_patched_obj_metadata(
...
         # Therefore, in the event of a key conflict, the value registered by the user-provided labels will take precedence.
-        total_metadata_size, labels, has_seen_keys = GCSObjectMetadataClient._add_labels_with_size_limitation(
-            normalized_custom_labels, total_metadata_size, max_gcs_metadata_size
-        )
-        _, labels, _ = GCSObjectMetadataClient._add_labels_with_size_limitation(
-            normalized_task_params_labels, total_metadata_size, max_gcs_metadata_size, labels, has_seen_keys
-        )
-        return dict(metadata) | dict(labels)
+        _merged_labels = GCSObjectMetadataClient._merge_custom_labels_and_task_params_labels(normalized_task_params_labels, normalized_custom_labels)
+        return dict(metadata) | dict(GCSObjectMetadataClient._adjust_gcs_metadata_limit_size(_merged_labels))
```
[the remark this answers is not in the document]
TlexCypher: Ahh... sorry, try to recontribute.

PR #450, opened 2025-03-05
Support functionalities to enhance task traceability with metadata for dependency search.

TlexCypher:
## Related works
https://github.com/m3dev/gokart/pull/445
https://github.com/m3dev/gokart/pull/446
https://github.com/m3dev/gokart/pull/448

## What does PR do?
In this Pull Request, I implement a metadata attribution feature that enables searching for tasks dependent on specific tasks executed with a given parameter set.

## Why is this needed?
Gokart caches the execution results and parameter states of each task in GCS. As shown in the Related Works section, various metadata are attached to each GCS object to enhance traceability. A common use case is searching for tasks that depend on a specific task executed with a given parameter set. Currently, Gokart does not support searching and tracing task dependencies from GCS metadata. This PR introduces this functionality.

## Pre-Requisists
The focus of this PR is embedding the necessary metadata to allow the CLI to search for specific dependencies. The search functionality itself will be implemented on the CLI side (CLI: https://github.com/TlexCypher/gcs-metadog).

## Checklist
CI is passing
Code formatting follows project standards.
Necessary tests have been added.
Existing tests pass.


gokart/utils.py
```diff
@@ -71,6 +71,19 @@ def flatten(targets: FlattenableItems[T]) -> list[T]:
     return flat
 
 
+K = TypeVar('K')
+
+
+def map_flattenable_items(items: FlattenableItems[T], func: Callable[[T], K]) -> FlattenableItems[K]:
```
TlexCypher, line 77: If we can use both Generics and isinstance at the same time, code would be below.
```python
def map_flattenable_items(items: FlattenableItems[T], func: Callable[[T], K]) -> FlattenableItems[K]:
    if isinstance(items, dict):
        return {k: map_flattenable_items(v, func) for k, v in items.items()}
    if isinstance(str):
        return items
    if isinstance(items, Iterable[T]):
        return [map_flattenable_items(i, func) for i in items]
    return func(items)
```

gokart/task.py
```diff
@@ -362,11 +365,33 @@ def dump(self, obj: Any, target: Union[None, str, TargetOnKart] = None, custom_l
...
+        )
+        required_task_outputs: dict[str, str] | None = None
+        if isinstance(_required_task_outputs, list):
+            required_task_outputs = {r.task_name: r.output_path for r in _required_task_outputs}
+        elif isinstance(_required_task_outputs, dict):
+            required_task_outputs = _required_task_outputs
+        else:
+            required_task_outputs = (
+                {_required_task_outputs.task_name: _required_task_outputs.output_path} if isinstance(_required_task_outputs, _RequiredTaskOutput) else None
+            )
```
mamo3gr, line 387: [imo]  
It would become more readable with extracting this section into a method, which returns `required_task_outputs`.

gokart/target.py
```diff
@@ -30,13 +30,23 @@ def exists(self) -> bool:
...
         return wrap_load_with_lock(func=self._load, task_lock_params=self._get_task_lock_params())()
 
-    def dump(self, obj, lock_at_dump: bool = True, task_params: dict[str, str] | None = None, custom_labels: dict[str, Any] | None = None) -> None:
+    def dump(
+        self,
+        obj,
+        lock_at_dump: bool = True,
+        task_params: dict[str, str] | None = None,
+        custom_labels: dict[str, Any] | None = None,
+        required_task_outputs: dict[str, str] | None = None,
```
mamo3gr, line 39: [imo]  
This parameter seems to be just a metadata. But its name may indicate that it effects the functionality of the method or the class's attribute. It would be better to rename for avoiding such a misleading.

mamo3gr:
LGTM  
I made some comments for improving code readability and leave them to your own choice to apply.

TlexCypher:
@mamo3gr Thank you for your thoughtful comments. I'm gonna deal with all of them.

gokart/utils.py
```diff
@@ -71,6 +71,19 @@ def flatten(targets: FlattenableItems[T]) -> list[T]:
...
+    if isinstance(items, dict):
+        return {k: map_flattenable_items(v, func) for k, v in items.items()}
+    if isinstance(items, str):
+        return items  # type: ignore
```
kitagry, line 81: In this case, `T` means `str, so you should apply func for this.

```suggestion
    if isinstance(items, str):
        return func(items) # type: ignore
```

gokart/utils.py
kitagry, line 77: https://docs.python.org/3.13/library/functions.html#map

python original map define `map(function, iterable)`, so you must suit python's manner. 

```suggestion
def map_flattenable_items(func: Callable[[T], K], items: FlattenableItems[T]) -> FlattenableItems[K]:
```

gokart/utils.py
```diff
@@ -71,6 +71,19 @@ def flatten(targets: FlattenableItems[T]) -> list[T]:
...
+    if isinstance(items, Iterable):
+        return [map_flattenable_items(i, func) for i in items]
```
kitagry, line 83: When pass `tuple[T]`, it should returns `tuple[K]`. But, in this implementation, this case is not cared for.

kitagry: And, could you add testcase?

gokart/gcs_obj_metadata_client.py
```diff
@@ -101,23 +104,28 @@ def _get_patched_obj_metadata(
...
-        for label_name, label_value in normalized_task_params.items():
-            if len(label_value) == 0:
-                logger.warning(f'value of label_name={label_name} is empty. So skip to add as a metadata.')
-                continue
-            if label_name in merged_labels.keys():
-                logger.warning(f'label_name={label_name} is already seen. So skip to add as a metadata.')
-                continue
-            merged_labels[label_name] = label_value
+        merged_labels: dict[str, str] = {}
+        for normalized_label in normalized_labels_list[:]:
```
kitagry, line 120:
```suggestion
        for normalized_label in normalized_labels_list:
```

kitagry (changes requested):

gokart/task.py
```diff
@@ -370,22 +370,15 @@ class _RequiredTaskOutput:
...
-            required_task_outputs = {r.task_name: r.output_path for r in _required_task_outputs}
-        elif isinstance(_required_task_outputs, dict):
-            required_task_outputs = _required_task_outputs
-        else:
-            required_task_outputs = (
-                {_required_task_outputs.task_name: _required_task_outputs.output_path} if isinstance(_required_task_outputs, _RequiredTaskOutput) else None
+        _required_task_outputs = flatten(
+            map_flattenable_items(
+                lambda task: map_flattenable_items(
+                    lambda output: _RequiredTaskOutput(task_name=task.get_task_family(), output_path=output.path()), task.output()
```
kitagry, line 376: output() returns `FlattenableItems[TargetOnKart]`, so I think you should use map_flattenable_items to extract output path.

gokart/gcs_obj_metadata_client.py
```diff
@@ -101,23 +106,51 @@ def _get_patched_obj_metadata(
...
+                else:
+                    flattened_list.append(item)
+            return flattened_list
+
+        if isinstance(required_task_outputs, dict):
+            return {k: GCSObjectMetadataClient._get_serialized_string(v) for k, v in required_task_outputs.items()}
+        if isinstance(required_task_outputs, tuple):
+            return tuple(required_task_output.serialize() for required_task_output in required_task_outputs)
+        if isinstance(required_task_outputs, Iterable):
+            return _list_flatten([GCSObjectMetadataClient._get_serialized_string(ro) for ro in required_task_outputs])
```
kitagry, line 137: tuple is one type of Iterable, and python has many iterable type, e.x) list, set, tuple, so on. You can write better like the following.

```suggestion
        if isinstance(required_task_outputs, Iterable):
            iter_type = type(required_task_outputs)
            return iter_type(_list_flatten([GCSObjectMetadataClient._get_serialized_string(ro) for ro in required_task_outputs]))
```

gokart/task.py
TlexCypher: @kitagry What do you mean?
As you said, output() returns `FlattenableItems[TargetOnKart]`, then what we wanna do is to retrieve only task family, task output path.

kitagry: Sorry, this comment is wrong! you are correct.

gokart/gcs_obj_metadata_client.py
kitagry: Sorry for the late reply.

Yes, you are correct, and you can change iterable to list,

And, What I actually meant was regarding the utils.map_flattenable_items method. I'll make another comment.

gokart/utils.py
```diff
@@ -72,6 +72,21 @@ def flatten(targets: FlattenableItems[T]) -> list[T]:
...
+
+def map_flattenable_items(func: Callable[[T], K], items: FlattenableItems[T]) -> FlattenableItems[K]:
+    if isinstance(items, dict):
+        return {k: map_flattenable_items(func, v) for k, v in items.items()}
+    if isinstance(items, tuple):
+        return tuple(map_flattenable_items(func, i) for i in items)
+    if isinstance(items, str):
+        return func(items)  # type: ignore
+    if isinstance(items, Iterable):
+        return [map_flattenable_items(func, i) for i in items]
```
kitagry, line 86:
```suggestion
        return map(lambda item: map_flattenable_items(func, i), items)
```

gokart/gcs_obj_metadata_client.py
TlexCypher: Thank you for replying.
I deal with your suggested changes.

gokart/task.py
```diff
@@ -5,12 +5,21 @@
...
 import types
 from collections.abc import Generator, Iterable
 from importlib import import_module
 from logging import getLogger
 from typing import Any, Callable, Generic, TypeVar, overload
 
+from gokart.required_task_output import RequiredTaskOutput
+from gokart.utils import map_flattenable_items
+
+if sys.version_info < (3, 13):
```
mski-iksm, line 18: Maybe this part is not needed?

gokart/gcs_obj_metadata_client.py
```diff
@@ -101,23 +107,49 @@ def _get_patched_obj_metadata(
...
-                logger.warning(f'value of label_name={label_name} is empty. So skip to add as a metadata.')
-                continue
-            if label_name in merged_labels.keys():
-                logger.warning(f'label_name={label_name} is already seen. So skip to add as a metadata.')
-                continue
-            merged_labels[label_name] = label_value
+        merged_labels: dict[str, str] = {}
+        for normalized_label in normalized_labels_list[:]:
+            for label_name, label_value in normalized_label.items():
+                if len(label_value) == 0:
```
mski-iksm, line 146: [MUST] This code may fail, since it seems to assume that `label_value ` is str.

I prefer checking if it is str, and then check the length as,

```
isinstance(label_value, str) and len(label_value)==0
```

gokart/gcs_obj_metadata_client.py
```diff
@@ -101,23 +107,49 @@ def _get_patched_obj_metadata(
...
-        for label_name, label_value in normalized_task_params.items():
-            if len(label_value) == 0:
-                logger.warning(f'value of label_name={label_name} is empty. So skip to add as a metadata.')
-                continue
-            if label_name in merged_labels.keys():
-                logger.warning(f'label_name={label_name} is already seen. So skip to add as a metadata.')
-                continue
-            merged_labels[label_name] = label_value
+        merged_labels: dict[str, str] = {}
+        for normalized_label in normalized_labels_list[:]:
```
mski-iksm, line 144: [weak-IMO]

```
for normalized_label in normalized_labels_list:
    for label_name, label_value in normalized_label.items():
        if len(label_value) == 0:
```

I thought this part a bit difficult to understand, since it is deeply nested.

It may get better if you extract `for label_name, label_value in...` part as a separate function, and apply it with a `functools.reduce()`.

However, current code is OK though. :)

mski-iksm (commented):
@TlexCypher 
I've made some comments but mainly LGTM! Thank you for your contribution!

gokart/gcs_obj_metadata_client.py
mski-iksm: @TlexCypher 
Then maybe the input `normalized_labels_list: list[dict[str, Any]]` should be `normalized_labels_list: list[dict[str, str]]` ?

gokart/gcs_obj_metadata_client.py
```diff
@@ -101,23 +107,44 @@ def _get_patched_obj_metadata(
...
+    @staticmethod
+    def _get_serialized_string(required_task_outputs: FlattenableItems[RequiredTaskOutput]) -> FlattenableItems[str]:
+        def _iterable_flatten(nested_list: Iterable) -> list[str]:
+            flattened_list: list[str] = []
+            for item in nested_list:
+                if isinstance(item, Iterable):
+                    flattened_list.extend(_iterable_flatten(item))
+                else:
+                    flattened_list.append(item)
+            return flattened_list
```
hirosassa, line 126: How about using `Iterator` like below. This can be (maybe) reduce temporal memory usage and it looks a bit elegant.
```suggestion
        def _iterable_flatten(nested_list: Iterable) -> Iterator[str]:
            for item in nested_list:
                if isinstance(item, Iterable):
                    yield from _iterable_flatten(item)
                else:
                    yield item
```

and we should change L130 as
```python
return list(_iterable_flatten([GCSObjectMetadataClient._get_serialized_string(ro) for ro in required_task_outputs]))
```

hirosassa (approved):
Commented!

gokart/gcs_obj_metadata_client.py
TlexCypher: You're right.
I put some changes to use `functools.reduce`.
Thank you for help!

gokart/gcs_obj_metadata_client.py
TlexCypher: Your suggestion is so efficient about memory usage!
I'm not familiar with `yield` and `yield from`, but this only returns generator.

In the original code (the one that returns a list), even with deep nesting or a large number of elements, it tends to be inefficient because it builds and holds the entire flattened list in memory at the end, and also creates intermediate lists during the recursive calls.

This approach is better because it's more memory-efficient, especially for large datasets or deeply nested structures, as it avoids keeping the entire flattened sequence in memory simultaneously until it's all collected.

I defenitely accept this suggest, thank you!

hirosassa (approved):
Looks good!

hirosassa:
@kitagry Hey, could you check this PR again?

gokart/gcs_obj_metadata_client.py
kitagry: I fixed here  3a60c45

PR #451, opened 2025-03-05
Identify abortion with dumping empty DataFrame

mamo3gr:
## Objective

When a task failed and raised an exception, we want to identify whether its reason is dumping an empty DataFrame (with `fail_on_empty_dump` option) or the others.

## Implementation

Raise specific exception for that case.

### DIFF
--- a/gokart/task.py
+++ b/gokart/task.py
@@ -38,6 +38,11 @@ T = TypeVar('T')
 K = TypeVar('K')
 
 
+# NOTE: inherited from AssertionError for backward compatibility (Formerly, Gokart raises that exception when a task dumps an empty DataFrame).
+class EmptyDumpError(AssertionError):
+    """Attempted to dump an empty DataFrame even though it is prohibited (fail_on_empty_dump is set to True)."""
+
+
 class TaskOnKart(luigi.Task, Generic[T]):
     """
     This is a wrapper class of luigi.Task.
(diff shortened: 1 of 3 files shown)


gokart/task.py
```diff
@@ -359,8 +364,9 @@ def dump(self, obj: Any, target: Union[str, TargetOnKart], custom_labels: dict[A
 
     def dump(self, obj: Any, target: Union[None, str, TargetOnKart] = None, custom_labels: dict[str, Any] | None = None) -> None:
         PandasTypeConfigMap().check(obj, task_namespace=self.task_namespace)
-        if self.fail_on_empty_dump and isinstance(obj, pd.DataFrame):
-            assert not obj.empty
+        if self.fail_on_empty_dump:
+            if isinstance(obj, pd.DataFrame) and obj.empty:
```
hirosassa, line 368: nits
```suggestion
        if self.fail_on_empty_dump and isinstance(obj, pd.DataFrame) and obj.empty:
```

hirosassa (approved):
added a comment but LGTM

gokart/task.py
mamo3gr: @hirosassa Even though these conditions can be merged as you mentioned, I intentionally split two `if` clauses; The first one focuses on the task's setting and the second one performs validation about empty.  
If any functionality would be appended there in the future, the first clause would not need to be modified.  
How does that sound?

kitagry (approved):
Thank you!!

PR #459, opened 2025-03-14
add lines parameter to JsonFileProcessor

kitagry:
I think auto enabled lines is a bit confusing. So, I think it is preferable to specify `lines=True` explicitly.


kitagry:
@ohyan Hi, Could you review this?

PR #470, opened 2025-11-24
drop python 3.9

hirosassa:
SSIA

FYI:
I tried adding Python 3.14 but some of dependencies are not followed currently.


kitagry (approved):
Thank you!

PR #471, opened 2025-12-14
feat: add automatic processor selection based on DataFrame type parameter

kitagry:
## Summary

This PR adds support for polars DataFrames alongside pandas DataFrames in gokart's file processors. Ta
sks can now specify their DataFrame type via the `TaskOnKart[T]` type parameter, and the appropriate pro
cessor will be automatically selected.

## Features

- **Auto-detection**: Automatically detects DataFrame type from `TaskOnKart[pd.DataFrame]` or `TaskOnK
art[pl.DataFrame]` type parameters
- **Backward compatible**: Defaults to pandas when no type parameter is specified

## Usage Example

```python
import polars as pl
from gokart import TaskOnKart

class MyPolarsTask(TaskOnKart[pl.DataFrame]):
    def output(self):
        return self.make_target('path/to/target.feather')

    def run(self):
        df = pl.DataFrame({'a': [1, 2, 3]})
        self.dump(df)  # Automatically uses polars-compatible processor

class MyPandasTask(TaskOnKart[pd.DataFrame]):
    def output(self):
        return self.make_target('path/to/target.feather')

    def run(self):
        df = pd.DataFrame({'a': [1, 2, 3]})
        self.dump(df)  # Uses pandas processor (default behavior)
```

### DIFF
--- a/gokart/file_processor/__init__.py
+++ b/gokart/file_processor/__init__.py
@@ -161,7 +161,7 @@ class FeatherFileProcessor(FileProcessor):
         return self._impl.dump(obj, file)
 
 
-def make_file_processor(file_path: str, dataframe_type: DataFrameType = 'pandas', store_index_in_feather: bool = True) -> FileProcessor:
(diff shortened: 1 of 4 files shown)


hirosassa (approved):
LGTM!

PR #457, opened 2025-03-10
Introduce `Polars` for dumping and loading data

hirosassa:
fixes https://github.com/m3dev/gokart/issues/304

In this PR, I introduced `polars` into gokart.
Currently the implementation is super rough and I didn't check the correctness.
Feel free to leave your comment for this software architecture / implementation.

PR #472, opened 2025-12-15
feat: add polars file processor

kitagry:
## Summary

This PR adds support for polars DataFrames alongside pandas DataFrames in gokart's file processors. This PR is very minimum support for Polars compared to #457 and #471 

## Features

- **Multi-backend support**: File processors now support both pandas and polars DataFrames
  - `CsvFileProcessor`
  - `JsonFileProcessor`
  - `ParquetFileProcessor`
  - `FeatherFileProcessor`
- **Backward compatible**: Defaults to pandas when no type parameter is specified

## Usage Example

```python
import polars as pl
from gokart import TaskOnKart

class MyPolarsTask(TaskOnKart):
    def output(self):
        return self.make_target('path/to/target.feather', processor=FeatherFileProcessor(dataframe_type='polars'))

    def run(self):
        df = pl.DataFrame({'a': [1, 2, 3]})
        self.dump(df)  # uses polars-compatible processor

class MyPandasTask(TaskOnKart):
    def output(self):
        return self.make_target('path/to/target.feather')

    def run(self):
        df = pd.DataFrame({'a': [1, 2, 3]})
        self.dump(df)  # Uses pandas processor (default behavior)
```

## Why not #457 ?

In #457, we switch between Polars and Pandas using GOKART_DATAFRAME_FRAMEWORK. However, this implies that we cannot use both at the same time. Since projects often migrate from Pandas gradually, we should allow users to use both Polars and Pandas simultaneously.

**This PR support minimum polars features, and then we will implement other useful features!**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,7 +14,7 @@ jobs:
         platform: ["ubuntu-latest"]
         tox-env: ["py310", "py311", "py312", "py313"]
         include:
-          - platform: macos-13
+          - platform: macos-15
             tox-env: "py313"
           - platform: macos-latest
             tox-env: "py313"


kitagry:
@hiro-o918 I fixed. Could you review again?

PR #504, opened 2026-04-08
move gcs and s3 deps to extras

hirosassa:
## Summary

- Move S3 (boto3) and GCS (google-auth, google-api-python-client) dependencies to optional extras, installable via pip install gokart[s3], pip install gokart[gcs], or pip install gokart[all]
- Remove unused dependency uritemplate
- Lazy-load S3/GCS-related imports, raising a clear ImportError with installation instructions when extras are missing

## Motivation

pip install gokart previously bundled cloud storage libraries even when unused, increasing package size unnecessarily.

- S3 extras (boto3 + transitive deps)
- GCS extras (google-api-python-client + transitive deps)

By making these optional, users who don't need S3/GCS can reduce dependencies by up to ~109 MB.

## Dependency size impact

Compared to `gokart[all]` (~110 MB of extras dependencies):

| Install command | Extras included | Savings vs `[all]` |
|---|---|---|
| `pip install gokart` | none | -110 MB |
| `pip install gokart[s3]` | S3 only (~24 MB) | -86 MB |
| `pip install gokart[gcs]` | GCS only (~86 MB) | -24 MB |
| `pip install gokart[all]` | both (baseline) | — |

## Breaking Changes

S3/GCS functionality is no longer available with a bare pip install gokart. Users must opt in:

```shell
pip install gokart[s3]   # S3 only
pip install gokart[gcs]  # GCS only
pip install gokart[all]  # both
```

## Changes

- pyproject.toml: Move boto3, google-auth, google-api-python-client from dependencies to optional-dependencies.
- s3_config.py, gcs_config.py, object_storage.py, gcs_obj_metadata_client.py, s3_zip_client.py, gcs_zip_client.py: Convert top-level imports to lazy imports with helpful error messages.
- test/test_extras.py: Add tests verifying behavior with and without extras installed.


hiro-o918 (approved):
LGTM!

gokart/s3_config.py
```diff
@@ -12,12 +11,16 @@ class S3Config(luigi.Config):
 
     _client = None
 
-    def get_s3_client(self) -> luigi.contrib.s3.S3Client:
+    def get_s3_client(self):
```
kitagry: Could you add something like the following? I think it would be better to add type annotations.

```python
if TYPE_CHECKING:
    import luigi.contrib.s3
```

kitagry (approved):
Sorry for the late review 🙏

gokart/s3_config.py
hirosassa: fixed in [7307a60](https://github.com/m3dev/gokart/pull/504/commits/7307a6030ce114b050b1d03072f06e14476ede2e)

hirosassa:
@hiro-o918 @kitagry Cloud you please take a look at changes in 077d5ab?
In this change, check actual dependency packages (boto3, googleapiclient) instead of luigi contrib modules to detect missing extras because luigi imports succeed without them and only fails at runtime.
This feels like reaching into luigi's internals, but there's no clean solution, I think.

kitagry (approved):

PR #508, opened 2026-05-10
fix: parameter name mismatch

hirosassa:
## What I did

Rename method parameters in `gokart/parameter.py` to match their parent class signatures

- `parse(self, s)` → `parse(self, x)` to match `luigi.Parameter.parse`
- `serialize(self, dt)` / `normalize(self, dt)` → `(self, x)` to match `luigi.Parameter`
- `default(self, obj)` → `default(self, o)` to match `json.JSONEncoder.default`

## Motivation

These mismatches cause `invalid-method-override` / `reportIncompatibleMethodOverride` errors in type checkers other than mypy (e.g. ty, Pyright), and can cause `TypeError` at runtime when callers use keyword arguments through a base class reference.


hiro-o918 (approved):

hiro-o918:
LGTM!

kitagry (approved):

hirosassa:
Thank you for your rapid review responses!

PR #511, opened 2026-05-10
Improve type definition of FlattenableItems for better type narrowing

hirosassa:
## What I did

Replace `Iterable` with concrete collection types (`list`, `tuple`) in the `FlattenableItems` recursive type alias to enable proper type narrowing after `isinstance` checks.

```python
# before
FlattenableItems = T | Iterable[FlattenableItems[T]] | dict[str, FlattenableItems[T]]

# after
FlattenableItems = T | list[FlattenableItems[T]] | tuple[FlattenableItems[T], ...] | dict[str, FlattenableItems[T]]
```
Since dict, list, and tuple are disjoint, type checkers can now correctly narrow the type after isinstance(x, dict), resolving many false positives in Pyright and other modern type checker tools.

## Motivation
The original Iterable union prevented type narrowing because dict is a subtype of Iterable, making the branches overlap. 
This caused Pyright to report errors on dict indexing, return types, and argument types across task.py, utils.py, and gcs_obj_metadata_client.py.

### DIFF
--- a/gokart/utils.py
+++ b/gokart/utils.py
@@ -27,7 +27,7 @@ def add_config(file_path: str) -> None:
 
 
 T = TypeVar('T')
-FlattenableItems: TypeAlias = T | Iterable['FlattenableItems[T]'] | dict[str, 'FlattenableItems[T]']
+FlattenableItems: TypeAlias = T | list['FlattenableItems[T]'] | tuple['FlattenableItems[T]', ...] | dict[str, 'FlattenableItems[T]']
 
 
 def flatten(targets: FlattenableItems[T]) -> list[T]:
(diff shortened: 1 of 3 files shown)


gokart/utils.py
```diff
@@ -76,7 +76,7 @@ def map_flattenable_items(func: Callable[[T], K], items: FlattenableItems[T]) ->
         return tuple(map_flattenable_items(func, i) for i in items)
     if isinstance(items, str):
         return func(items)  # type: ignore
-    if isinstance(items, Iterable):
+    if isinstance(items, list):
```
hiro-o918, line 79: :memo: 
tuple is covered by L75

hiro-o918 (approved):

PR #517, opened 2026-05-13
add casts for pandas type stub incompatibilities

hirosassa:
SSIA

### DIFF
--- a/gokart/utils.py
+++ b/gokart/utils.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import os
 from collections.abc import Callable, Iterable
 from io import BytesIO
-from typing import Any, Literal, Protocol, TypeAlias, TypeVar, get_args, get_origin
+from typing import Any, Literal, Protocol, TypeAlias, TypeVar, cast, get_args, get_origin
 
 import dill
 import luigi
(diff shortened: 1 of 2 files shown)


hiro-o918:
Regarding `return pd.read_pickle(cast(Any, file))` in `gokart/utils.py`:

The `cast(Any, ...)` here could be eliminated by refining the `FileLike` Protocol so that it becomes a structural subtype of pandas' `ReadPickleBuffer` (adding `mode` property, `tell()`, and adjusting `read` / `seek` signatures). That would also require updating `_ChunkedLargeFileReader` in `gokart/file_processor/base.py` to match.

Since that's a wider change, let's keep this PR focused on the cast workaround and track the refactor separately: #521.

hiro-o918 (approved):

hirosassa:
I agree with your suggestion. Will apply your idea in another PR.

gokart/utils.py
```diff
@@ -91,7 +91,7 @@ def load_dill_with_pandas_backward_compatibility(file: FileLike | BytesIO) -> An
     except Exception:
         assert file.seekable(), f'{file} is not seekable.'
         file.seek(0)
-        return pd.read_pickle(file)
+        return pd.read_pickle(cast(Any, file))
```
hiro-o918, line 94: ah, how about casting to `ReadPickleBuffer` of pandas because some method of pandas protocol seems to be too much?

hirosassa: agreed, will apply in another PR.

PR #518, opened 2026-05-13
fix type issues in config_params.py and run.py

hirosassa:
SSIA

### DIFF
--- a/gokart/config_params.py
+++ b/gokart/config_params.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Any
+from typing import Any, cast
 
 import luigi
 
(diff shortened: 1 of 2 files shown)


gokart/config_params.py
```diff
@@ -27,10 +27,10 @@ class Wrapped(task_class):  # type: ignore
             @classmethod
             def get_param_values(cls, params, args, kwargs):
                 for param_key, param_value in self._config_class().param_kwargs.items():
-                    task_param_key = self._parameter_alias.get(param_key, param_key)
+                    task_param_key = self._parameter_alias.get(param_key) or param_key
```
hiro-o918: [nits]
I wonder if we actually need this change.

hiro-o918 (approved):

gokart/config_params.py
hirosassa: Thanks! It's unecessary. fixed in [df762bc](https://github.com/m3dev/gokart/pull/518/commits/df762bc50ab20b1d69e626f299ccc6af160b2576)

PR #525, opened 2026-06-09
Resolve DataFrame type through intermediate generic classes

yusuke14tanaka:
## Problem

`get_dataframe_type_from_task` infers the DataFrame backend (pandas / polars / polars-lazy) from the `TaskOnKart[T]` type parameter to choose the matching file processor. It inspects the type argument of a `TaskOnKart[...]` base directly, so when the concrete type is bound through an *intermediate generic class*, the `TaskOnKart` base still carries an unbound `TypeVar` and detection silently falls back to pandas.

This shows up with a common pattern where a library defines a generic base task and the concrete type is bound on a subclass:

```python
T = TypeVar('T')

class MyBase(TaskOnKart[T], Generic[T]):
    ...

class MyTask(MyBase[pl.DataFrame]):  # detected as pandas before this change
    ...
```

A polars task detected as pandas means its cache is read/written with the pandas processor.

## What this changes

Walk the MRO to collect `TypeVar` → type-argument bindings, then resolve the `TaskOnKart` type argument through that map (following `TypeVar` → `TypeVar` chains across multiple generic levels). Direct and nested cases are unaffected, since a non-`TypeVar` argument resolves to itself.

Added tests cover the intermediate-generic case for both `pl.DataFrame` and `pl.LazyFrame`.

### DIFF
--- a/gokart/utils.py
+++ b/gokart/utils.py
@@ -94,6 +94,16 @@ def load_dill_with_pandas_backward_compatibility(file: FileLike | BytesIO) -> An
         return pd.read_pickle(cast(Any, file))
 
 
+def _resolve_type_var(type_arg: Any, substitutions: dict[TypeVar, Any]) -> Any:
+    # Follow a chain of TypeVar substitutions (e.g. T_a -> T_b -> pl.DataFrame) to a concrete type.
+    # seen guards against cyclic substitutions.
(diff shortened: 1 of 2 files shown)


kitagry (approved):
Thank you!

hirosassa (approved):
LGTM

PR #530, opened 2026-07-21
fix: require pandas>=2.3.3 on Python 3.14

hiro-o918:
## Background

Investigating why only the `tests (ubuntu-latest, py314)` job failed on #529 revealed the root cause is pandas, not pillow.

- The failure is a segmentation fault inside pandas' datetime C extension at `test/test_pandas_type_config.py::test_datetime_success`, deterministically reproducible on re-run: https://github.com/m3dev/gokart/actions/runs/29820432602
- pandas 2.2.3 ships no cp314 wheels, so on Python 3.14 it is always built from sdist — an unsupported, untested configuration (the first pandas release with cp314 wheels is 2.3.3)
- A full diff of all 82 installed packages between the last green master run and the failing run shows the only change is pillow 12.2.0 → 12.3.0. Pillow 12.3.0's C-extension optimizations changed the process memory layout, which exposed latent undefined behavior in the sdist-built pandas — pillow is the trigger, pandas is the actual defect.

## Changes

- `pyproject.toml`: split the pandas requirement with environment markers — unchanged (unconstrained) for `python_version < '3.14'`, `pandas>=2.3.3` for `python_version >= '3.14'`
- `uv.lock`: regenerated. pandas is forked to 2.3.3 (cp314 wheels) for Python >= 3.14 while staying at 2.2.3 for <= 3.13. No other package versions changed (only mechanical resolution-marker splits).

## QA

- `uv sync --locked -p 3.14 --group test` installs pandas 2.3.3 from a prebuilt cp314 wheel (`Tag: cp314-cp314-macosx_11_0_arm64`)
- `test/test_pandas_type_config.py` (previously segfaulting) passes all 6 tests on Python 3.14
- `uv sync --locked -p 3.12` still resolves pandas 2.2.3 (resolution for <= 3.13 untouched); the same tests pass
- `uv run ruff check .` / `uv run ruff format --check .` / `uv run mypy gokart test` all pass
- Marker boundary conditions evaluated with `packaging.markers.Marker.evaluate()` (3.13.9 → old branch; 3.14.0rc1 / 3.14.0 / 3.15.0 → new branch; mutually exclusive and exhaustive)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,9 @@ requires-python = ">=3.10, <4"
 dependencies = [
   "luigi>=3.8.1",  # 3.8.1+ required for ConfigPath TypedDict and OptionalParameterMixin generics
   "slack-sdk",
-  "pandas",
+  # pandas < 2.3.3 has no cp314 wheels; the sdist-built C extension segfaults on Python 3.14
+  "pandas; python_version<'3.14'",
+  "pandas>=2.3.3; python_version>='3.14'",
   "numpy",
   "pyarrow",
   "APScheduler",
--- a/uv.lock
+++ b/uv.lock
@@ -2,8 +2,10 @@ version = 1
 revision = 3
 requires-python = ">=3.10, <4"
 resolution-markers = [
-    "python_full_version >= '3.13' and sys_platform == 'win32'",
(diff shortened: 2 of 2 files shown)


hirosassa (approved):
LGTM!

